PR #116, opened 2026-04-10
style: apply ruff format + fix pre-existing lint errors (unblocks CI)

Mikecranesync:
## Summary

- Brings the repo into compliance with the CI Lint & Format gate (`ruff==0.9.*`). CI has been red on `main` due to accumulated pre-existing drift — **29 files** needed `ruff format`, **13 lint errors** needed `ruff check --fix` (12 auto-fixable, 1 manual).
- Unblocks PR #115 (`feat: set hnsw.ef_search=100 for recall query tuning`) and every future PR that touches the drifted files.
- Pure style/lint cleanup. **Zero behavior change** intended.

## What's in the diff

**`ruff format` (29 files)** — multi-line signature splitting, trailing commas, consistent line breaks. No semantic change. Paths touched:
- `mira-bots/shared/` — 9 files (`adapters/base.py`, `benchmark_db.py`, `engine.py`, `eval/evaluator.py`, `guardrails.py`, `inference/router.py`, `langfuse_setup.py`, `nemotron.py`, `neon_recall.py`, `tts.py`, `workers/print_worker.py`, `workers/rag_worker.py`, `workers/vision_worker.py`)
- `mira-bots/slack/` — `bot.py`, `pdf_handler.py`
- `mira-bots/telegram/` — `bot.py`
- `mira-core/mira-ingest/` — `db/neon.py`, `main.py`, `tests/test_benchmark.py`, `tests/test_ingest.py`, `tests/test_prejudged.py`
- `mira-mcp/` — `cmms/atlas.py`, `cmms/base.py`, `cmms/factory.py`, `cmms/fiix.py`, `cmms/limble.py`, `cmms/maintainx.py`, `context/viking_store.py`, `server.py`

**`ruff check --fix` (12 auto-fixed errors)**:
- `I001 import-block-sort` — `evaluator.py`, `slack/bot.py`, `telegram/bot.py`, `test_benchmark.py` (×2), `test_ingest.py`, `test_prejudged.py`
- `F401 unused-import` — `main.py` (`sqlite3`), `test_ingest.py` (`json`, `sqlite3`), `viking_store.py` (`math`)
- `F541 f-string-without-placeholders` — `evaluator.py`

**1 manual fix (F841, not auto-fixable)**:
[...]

PR #118, opened 2026-04-10
test: fix 4 stale tests exposed by CI lint cleanup (removes #116 skips)

Mikecranesync:
## Summary

Unwinds every skip marker added on [#116](https://github.com/Mikecranesync/MIRA/pull/116) and replaces them with real fixes. Four pre-existing broken tests, all exposed when #116 turned the Lint & Format gate green and let `test-unit` actually run for the first time in a while.

**Base:** `claude/ci-ruff-format-cleanup` (stacked on #116). Will auto-retarget to `main` when #116 merges.

**Commit order on this branch (top to bottom):**
1. `d799a72` — fix stale image_downscale + reranking tests (removes #116 skips)
2. `9417742` — rewrite reddit_harvest test to mock _fetch_search seam

## Why each test was broken (all pre-existing on `main`)

All four failures reproduce identically against `origin/main` with unmodified test files — verified by checking out `main`'s copy and running the same pytest commands. #116 didn't cause any of them; it merely turned the red lint gate green so `test-unit` stopped being skipped, which exposed the rot.

### 1. `test_benchmark.py::test_harvest_parses_json_and_filters`

`mira-core/scripts/reddit_harvest.py` was refactored to:
- route all Reddit fetches through a private `_fetch_search(sub, query, client)` helper that hits `/r/{sub}/search.json`, not the old top-level listing
- filter posts through a `DIAGNOSTIC_KEYWORDS` regex plus a question-shape check in `_is_relevant()`
- require `is_self=True` on each post
- throttle with `time.sleep(2)` between 5 subs × 6 queries = 30 searches

The existing test mocked `httpx.Client.get` at the class level — a seam `_fetch_search()` doesn't hit the way the test expects. The patch silently fell through and the test made **real** Reddit API calls, returning `0` harvested instead of the expected `2`. Plus: no `is_self=True`, no `selftext`, no `time.sleep` stub.

**Fix:** patch `reddit_harvest._fetch_search` directly (the single seam between `harvest()` and Reddit), plus `reddit_harvest.time.sleep` to keep the test fast (~1 s vs ~60 s). Updated fake fixture to include `is_self=True` + realistic `selftext`. Added a second assertion on `result["total"]` to anchor the dedup path.

### 2. `test_image_downscale.py::test_large_image_resized_to_512`

`_resize_for_vision()` was retuned to `MAX_VISION_PX=1024` (see `mira-bots/telegram/bot.py:97`) but the test still pinned the old 512 cap.

**Fix:** rename to `test_large_image_resized_to_vision_max_px` and read `MAX_VISION_PX` from the same env var the production code uses. The test now tracks tuning changes automatically instead of re-breaking every time someone tweaks the cap. Drops the now-unused `pytest` import.

### 4. `test_reranking.py::test_photo_query_embeds_with_asset_context`

Same root cause. `RAGWorker.process()` was refactored to add a `_visual_search()` dual-modality retrieval path: when `photo_b64` is set **and** `self._ingest_url` is set, photo queries route through an HTTP call to `mira-ingest/ingest/search-visual`, silently bypassing the `_neon_recall` mock these tests rely on. The mocks never fired; `_visual_search` 403'd to the real mira-ingest URL; `rerank` and `_embed_ollama` were never invoked; the assertions blew up.

**Fix:** set `worker._ingest_url = None` at the top of each test, forcing `process()` into the `else` branch (text-embed + `_neon_recall`) the tests were originally written against. Smallest change that preserves the tests' intent (verify rerank gets called on photo queries; verify the embed query includes asset context).

## Also on this branch: test_tts.py path fix

Came in via the rebase from #116 (`9c9dbc6`). `test_tts.py` imported from `mira-bots/telegram/tts.py`, but `tts.py` lives at `mira-bots/shared/tts.py` (refactor moved it without updating the test). One-line fix: sys.path insert now points at `../shared`. Without this fix, the entire `mira-bots/tests/` collection died fast at import time, which is why the CI Unit Tests job was finishing in ~21 s instead of ~80 s.

## Verification (all under CI-pinned `ruff==0.9.10`)

- [x] `pytest mira-core/mira-ingest/tests/test_benchmark.py::test_harvest_parses_json_and_filters` → **passed**
- [x] `pytest mira-bots/tests/test_image_downscale.py` → **3 passed**
- [x] `pytest mira-bots/tests/test_reranking.py` → **9 passed**
- [x] `pytest mira-bots/tests/ --ignore=mira-bots/tests/test_slack_relay.py` → **39 passed, 8 skipped** (was 36+11 on cleanup branch — the 3 temporary skips I added on #116 are now real passes)
- [x] `pytest mira-core/mira-ingest/tests/` → **28 passed** (was 27+1 on cleanup — reddit skip is now a real pass)
- [x] `ruff check <CI paths>` → **All checks passed**
- [x] `ruff format --check <CI paths>` → **42 files already formatted**
- [x] Zero `@pytest.mark.skip` markers remain from the cleanup sequence

## Test plan

- [ ] Merge [#116](https://github.com/Mikecranesync/MIRA/pull/116) first (the style cleanup + skip markers)
- [ ] GitHub auto-retargets this PR from `claude/ci-ruff-format-cleanup` to `main`
- [ ] CI Lint & Format + Unit Tests + License Check all green on this PR
- [ ] Merge this PR
- [ ] Rebase [#115](https://github.com/Mikecranesync/MIRA/pull/115) (`feat: set hnsw.ef_search=100`) onto new `main` — should go fully green in one shot

## Risks / caveats

1. **Scope creep from #118's original title.** Originally this PR only covered the Reddit test; now it covers 4 stale tests. Title and body updated to reflect.
2. **The `seen_post_ids` dedup in the reddit test is what makes the 90 mocked calls collapse to 2 inserts.** If a future refactor changes dedup semantics, the assertion will drift. Pinned via both `result["harvested"] == 2` and `result["total"] == 2` as anchors.
3. **The reranking tests still only cover the text-embed branch.** They don't exercise `_visual_search()` itself — that's a separate test that needs its own mock of the mira-ingest HTTP endpoint. Out of scope for this PR.
4. **`test_image_downscale` now depends on the MAX_VISION_PX env var.** If CI sets `MAX_VISION_PX` to something non-default, the test uses that value. This is intentional — the test should track production behavior, not a hardcoded snapshot.

https://claude.ai/code/session_016AceA9WFkgpF2dqXBwFGJw

### DIFF
--- a/mira-bots/tests/test_image_downscale.py
+++ b/mira-bots/tests/test_image_downscale.py
@@ -4,8 +4,6 @@ import io
 import os
 import sys
 
-import pytest
-
 # Set dummy env vars for bot.py import
 os.environ.setdefault("TELEGRAM_BOT_TOKEN", "dummy-token-for-testing")
 os.environ.setdefault("OPENWEBUI_BASE_URL",
@@ -22,20 +20,19 @@ from bot import _resize_for_vision
 from PIL import Image
 
 
-@pytest.mark.skip(
-    reason="Stale — _resize_for_vision() was retuned to MAX_VISION_PX=1024 "
-    "(see mira-bots/telegram/bot.py:97). Assertion still pins the old 512 "
-    "cap. Skipped to unblock CI as part of the style cleanup; real fix "
-    "updates the assertion to match the current 1024 target."
-)
-def test_large_image_resized_to_512():
-    """1920x1080 image should be downscaled so max side <= 512."""
+def test_large_image_resized_to_vision_max_px():
+    """1920x1080 image should be downscaled so max side <= MAX_VISION_PX.
+
+    Value tracks the MAX_VISION_PX env var (default 1024, see
+    mira-bots/telegram/bot.py:_resize_for_vision).
+    """
+    max_px = int(os.environ.get("MAX_VISION_PX", "1024"))
     img = Image.new("RGB", (1920, 1080), (128, 128, 128))
     buf = io.BytesIO()
     img.save(buf, "JPEG")
     result = _resize_for_vision(buf.getvalue())
     out = Image.open(io.BytesIO(result))
-    assert max(out.size) <= 512
+    assert max(out.size) <= max_px
 
 
 def test_small_image_unchanged():
--- a/mira-bots/tests/test_reranking.py
+++ b/mira-bots/tests/test_reranking.py
@@ -140,16 +140,15 @@ class TestRerankingIntegration:
         query_arg = worker.nemotron.rerank.call_args[0][0]
         assert "F004" in query_arg
 
-    @pytest.mark.skip(
-        reason="Stale — RAGWorker.process() photo-query path was refactored "
-        "to call _visual_search() when self._ingest_url is set, bypassing the "
-        "text-embed + _neon_recall path this test mocks. Skipped to unblock "
-        "CI as part of the style cleanup; real fix sets worker._ingest_url = "
-        "None to force the text-embed branch."
-    )
     @pytest.mark.asyncio
     async def test_rerank_called_for_photo_query(self):
         worker = _make_rag_worker()
+        # Force the text-embed + _neon_recall branch. When self._ingest_url is
+        # set, RAGWorker.process() routes photo queries through _visual_search()
+        # (dual-modality retrieval via HTTP to mira-ingest/ingest/search-visual)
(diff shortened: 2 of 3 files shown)


Mikecranesync:
Test fixes landed in pipeline branch.

PR #886, opened 2026-04-30
fix(agents): silence Telegram 400s + fix Atlas DNS for cron context

Mikecranesync:
## Summary

Two of the daily-cron agents have been pushing error alerts into Telegram every run. Both root causes are trivial and independent:

### 1. Telegram HTTP 400 swallowing alerts (`lead_scout`, `asset_intel`)

`orchestrator._send_telegram()` always uses `parse_mode=Markdown`. When an agent's result dict contains a value with an unbalanced `_`, `*`, or `[]` (a region name like `north_america`, or a model number containing `_`), Telegram returns 400 "can't parse entities" and the alert is dropped to the log. We see exactly this in `lead_scout.log` and `asset_intel.log` (single line: `Telegram send failed: HTTP Error 400: Bad Request`).

**Fix:** on HTTP 400, retry once with no `parse_mode`. Markdown rendering is nice-to-have; delivery isn't. Also collapsed the bespoke send block in `run_daily_digest()` to share the same helper, so the digest gets the same fallback.

### 2. Atlas DNS failure (`cmms_sync`, `pm_escalation`)

Both scripts read `ATLAS_API_URL` from Doppler, which is `http://cmms-backend:8080` (Docker DNS name) — for in-container callers. Cron runs on the host, where that name doesn't resolve. Every run fails:

```
httpx.ConnectError: [Errno -3] Temporary failure in name resolution
```

**Fix:** read a new `ATLAS_API_HOST_URL` env var first, default to `http://localhost:8082` (the actual published host port — confirmed via `docker port cmms-backend` and a `curl localhost:8082` that returned 403 "auth required", proving the service is up). Container callers' use of `ATLAS_API_URL` is untouched.

## Out of scope (surfaced separately to Mike)

- `kb_growth` Docling 504 on PowerFlex-525 — `DOCLING_SERVE_MAX_SYNC_WAIT=600` is already applied but a 4.5 MB PDF still times out. Needs different fix (chunked retry tuning) — flagged.
- `morning_brief` `ModuleNotFoundError` — historical; commit `fcf878c` already fixed the import. Next 09:00 UTC cron will confirm.
- `mira-pipeline` Ollama embed failures — Bravo Mac Mini at `100.86.236.11:11434` is unreachable. Not an agent issue; ops/host problem.
- `mira-web` Resend 429 + missing `blog_drafts` table — separate from the agent system.

## Test plan

- [x] `python3 -c "import ast; ast.parse(...)"` on all three files — clean
- [ ] After merge + VPS pull, next cron runs (CMMS sync 16:00 ET / 20:00 UTC, PM escalation 22:00 ET / 02:00 UTC) — expect green tracebacks gone
- [ ] Verify lead_scout/asset_intel send a real Telegram message (instead of failing 400) on next 10:00 ET / 18:00 ET runs

### DIFF
--- a/mira-bots/shared/chat/types.py
+++ b/mira-bots/shared/chat/types.py
@@ -31,7 +31,9 @@ class NormalizedChatEvent:
     """One inbound message from any platform, normalized."""
 
     event_id: str
-    platform: Literal["telegram", "slack", "teams", "gchat", "webui", "email", "whatsapp", "webchat"]
+    platform: Literal[
+        "telegram", "slack", "teams", "gchat", "webui", "email", "whatsapp", "webchat"
+    ]
     tenant_id: str
     user_id: str  # canonical MIRA user ID (after identity resolution)
     external_user_id: str  # platform-specific user ID
--- a/mira-bots/shared/engine.py
+++ b/mira-bots/shared/engine.py
@@ -620,9 +620,7 @@ class Supervisor:
             # Procedural how-to questions: answer from knowledge, skip doc crawl.
             # Keyword "instructional" also routes here via the fallback mapping above.
             if _router_intent == "answer_question" or _keyword_intent == "instructional":
-                return await self._handle_instructional_question(
-                    chat_id, message, state, trace_id
-                )
+                return await self._handle_instructional_question(chat_id, message, state, trace_id)
 
             # find_documentation: let the existing specificity-gate block handle it below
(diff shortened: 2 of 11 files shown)


Mikecranesync:
Linear mirror: https://linear.app/cranesync/issue/CRA-55 (filed during 2026-05-04 tracker reconciliation pass)

PR #642, opened 2026-04-25
feat(assets): Unit 5 — UNS-style asset model + /asset command

Mikecranesync:
## Summary

Closes **Unit 5** per [docs/plans/2026-04-19-mira-90-day-mvp.md](docs/plans/2026-04-19-mira-90-day-mvp.md). Foundation for hierarchical asset queries; unblocks Unit 7 (QR pre-load uses the same asset metadata).

- **Migration `011_assets_uns.sql`** — ltree extension + `assets` table with `uns_path ltree NOT NULL`, GIST index, tenant unique constraint
- **Read path `mira-bots/shared/assets.py`** — `list_top_levels` / `list_children` / `get_asset` helpers; fail-closed on missing DB or malformed input (matches `neon_recall.py` contract)
- **Telegram `/asset` command** — no-arg shows top 2 levels; `<path>` descends or shows asset card
- **Backfill `mira-core/scripts/backfill_uns.py`** — CSV → assets, idempotent ON CONFLICT, `--dry-run` mode
- **29 unit tests, all green** — path validation incl. SQL-injection attempts, formatter rendering, safety guards

## Migration numbering note

Plan said `005_assets_uns.sql`. Bumped to **011** because 004-010 were taken in flight (channel config, guest reports, knowledge tsvector for Unit 6, ingested files, hub user/tenant tables, tenant_id backfills). 011 is purely additive; no ordering dependency on prior migrations.

## Coordination

- Plan in-flight table updated (`2472cea`) before edits
- Pre-flight: no other branch/PR touches `mira-bots/shared/`, `mira-bots/telegram/bot.py`, or `mira-core/mira-ingest/db/migrations/`
- Branched from latest `origin/main` (`3816af0`, post Unit 9a merge)

## Test plan
- [x] `pytest mira-bots/tests/test_assets.py -v` — 29 passed
- [x] `python -m ruff check` on all touched files — clean
- [x] AST parses on bot.py / assets.py / backfill_uns.py
- [ ] **Live migration apply on NeonDB** — deferred to Mike's hand-run (not in PR scope)
- [ ] **DoD live demo** — seed fixture tenant w/ 4-level hierarchy (15 assets), `/asset enterprise.site_a` returns level-3 children — runs after migration applied

## Out of scope (refuse mid-execution per plan)
- Atlas → assets sync (that's Unit 8 territory)
- Multi-tenant signup flow for asset CSV upload (n=1 default)
- Real customer hierarchy mappings (post-pilot)

[...]

PR #887, opened 2026-04-30
fix: silence post-sweep noise — Resend 429 + blog_drafts + Docling fallback

Mikecranesync:
## Summary

Three independent log/UX issues from the cowork sweep, bundled into one PR because the diffs are small and isolated:

### 1. Resend 429 rate-limit loop (`mira-web/src/lib/mailer.ts`)

The drip dispatcher (`for tenant of tenants { await sendEmail() }`) and Apps-Script-driven inbox-receipt bursts both fire faster than Resend's 5/sec free-tier cap, producing the `[mailer] Resend API error: 429` lines we've been seeing in `mira-web` logs.

**Fix:** every Resend POST now goes through `resendFetch()`, which:
- Serializes through a chained Promise pacer with `MIN_SEND_INTERVAL_MS = 250` (= ≤4/sec, fits the cap with headroom)
- On 429, backs off based on the `Retry-After` header (capped 0.5–5 s) and retries once

Both `sendEmail()` and `sendInboxReceiptEmail()` now route through it. Global throttle, no caller changes.

### 2. `blog_drafts` missing-table error (`mira-web/src/lib/blog-db.ts`)

`getLiveFaultCodes()` and `getLiveBlogPosts()` query a `blog_drafts` table that doesn't exist yet — feature shipped without its migration. Each request was logging a NeonDbError. The catch already returned `[]` gracefully; the noise was just `console.error`.

**Fix:** detect Postgres SQLSTATE `42P01` (or the regex `relation "blog_drafts" does not exist`) and silently cache an empty result for `TTL_MS` so we don't re-query on every page load. When Mike provisions the editorial workflow, the table appears, the cache TTL expires, and live data flows in — no code change needed.

Trade-off considered: writing the migration now (`013_blog_drafts.sql`). Skipped because (a) the editorial workflow isn't designed yet, (b) PRs #642/#646 are already contending for migration numbers 011/012.

### 3. Docling 504 timeout on heavy-image PDFs (`mira-crawler/tasks/full_ingest_pipeline.py`)

`kb_growth` failed on Allen-Bradley PowerFlex-525 (4.5 MB, 38 pages) on 2026-04-29 even with `DOCLING_SERVE_MAX_SYNC_WAIT=600` already applied — every 15-page split chunk 504'd. Result: queue stuck, daily Telegram error.

**Fix:** when the 15-page split returns 0 chars, retry once with `chunk_pages=5` before giving up. Smaller chunks usually finish well under the 10-min timeout. Marked as method `split-5` in the report so it's distinguishable.

PowerFlex-525 was manually marked `done` in the queue overnight, so the immediate burn isn't there — but next time a heavy PDF lands, the queue keeps moving instead of stalling.

## Out of scope (handled separately)

- **Bravo Ollama lockdown** — done operationally (launchd plist now has `OLLAMA_HOST=100.86.236.11:11434`, persistent across reboot). No repo changes needed; production callers already used the Tailscale IP. Verified: localhost + LAN refused, Tailscale IP serves model list.
- **morning_brief import error** — was historical; `fcf878c` on `feat/asset-intelligence` already fixed it. Today's 09:00 UTC cron fired clean (syslog confirms). Manual re-run also exits 0.
- **kb_growth queue-advance** (so a permanently-failing PDF doesn't get retried daily forever) — needs to land on `feat/asset-intelligence` since `run_kb_growth.py` only exists there. Stacking onto PR #886 separately.

## Test plan

- [x] `python3 -m ast` parse on `full_ingest_pipeline.py` — clean
- [x] `bun build mailer.ts` — succeeds (`Bundled 1 module in 343ms`)
- [x] `bun build blog-db.ts` — only complains about missing `node_modules`, not the edit
- [ ] After merge + VPS rebuild of mira-web: confirm no `Resend API error: 429` and no `blog_drafts does not exist` in `docker logs mira-web --since 1h`
- [ ] Re-queue a heavy PDF in `mira-crawler/cron/manual_queue.json` and watch tomorrow's 06:00 UTC kb_growth — expect `split-5` method, non-empty chars


Mikecranesync:
Linear mirror: https://linear.app/cranesync/issue/CRA-56 (filed during 2026-05-04 tracker reconciliation pass)

PR #891, opened 2026-04-30
fix(inbox): P0.4 — per-IP rate limit on /api/v1/inbox/email (10 req/min)

Mikecranesync:
## Summary — P0.4 inbox rate limit

Audit found 0 throttling on `/api/v1/inbox/email` — verified live with 10 parallel POSTs, all returning 401 with no rate-limit kicking in. Each request runs HMAC-SHA256 over the raw body; sustained junk traffic burns CPU and floods the audit log.

**Scope narrowed:** `/api/register` turned out to already have `checkRegisterRateLimit()` (verified live: 5 then 429). The original P0.4 was overscoped against a clean codebase scan. This PR is inbox-only.

## What changed

- `mira-web/src/routes/inbox.ts` — added in-memory sliding-window rate limit:
  - **10 requests per 60s per source IP**
  - Returns `429` + `Retry-After: <seconds>` header
  - Gate runs **before** HMAC compute so a flood doesn't burn signature work
  - IP from `X-Forwarded-For` first hop, falling back to `X-Real-IP` and `CF-Connecting-IP`
  - Opportunistic GC keeps the map bounded under sustained junk traffic
- `_testing` export added so the unit tests can reset state and exercise the bucket
- New test file `inbox-rate-limit.test.ts` (9 tests) covers sliding window, boundary, recovery, per-IP isolation, and the IP-extraction header precedence
- Existing `inbox.test.ts` got `beforeEach(_testing.reset())` so the suite doesn't blow past the bucket on `ip="unknown"` (shared across all in-process `inbox.request()` calls)

## Why in-memory

`mira-web` is currently n=1 container. Single source of truth, no extra dep. The file comment flags that horizontal scale would lift this to the existing `infra_redis_1` Redis container — should be a 30-min refactor at that point, not blocking now.

## Apps Script behavior unchanged

Apps Script fires once per email (max ~1/min in observed practice). 10/min headroom comfortably covers a customer who forwards 5 PDFs at once and any retries the script does.

## Test plan

- [x] `bun test inbox.test.ts inbox-rate-limit.test.ts` → **36/36 pass**
- [ ] After merge: `for i in {1..15}; do curl -X POST https://factorylm.com/api/v1/inbox/email -H "X-Hmac-Signature: 00" -d '{}'; done` — expect first 10 to return 401, the rest to return 429 with Retry-After

### DIFF
--- a/mira-bots/shared/chat/types.py
+++ b/mira-bots/shared/chat/types.py
@@ -31,7 +31,9 @@ class NormalizedChatEvent:
     """One inbound message from any platform, normalized."""
 
     event_id: str
-    platform: Literal["telegram", "slack", "teams", "gchat", "webui", "email", "whatsapp", "webchat"]
+    platform: Literal[
+        "telegram", "slack", "teams", "gchat", "webui", "email", "whatsapp", "webchat"
+    ]
     tenant_id: str
     user_id: str  # canonical MIRA user ID (after identity resolution)
     external_user_id: str  # platform-specific user ID
--- a/mira-bots/shared/engine.py
+++ b/mira-bots/shared/engine.py
@@ -620,9 +620,7 @@ class Supervisor:
             # Procedural how-to questions: answer from knowledge, skip doc crawl.
             # Keyword "instructional" also routes here via the fallback mapping above.
             if _router_intent == "answer_question" or _keyword_intent == "instructional":
-                return await self._handle_instructional_question(
-                    chat_id, message, state, trace_id
-                )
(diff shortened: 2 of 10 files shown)


Mikecranesync:
Linear mirror: https://linear.app/cranesync/issue/CRA-53 (filed during 2026-05-04 tracker reconciliation pass)

Mikecranesync:
Superseded by the re-cut just opened (see PR referencing this one) — the inbox rate-limit hunks were ported cleanly onto current main; this stale branch also bundled unrelated bot-engine files that must not ride along. (Worklist #2990 item P0-3.)

PR #936, opened 2026-05-03
fix(web): register /images/* static route to unblock hero cartoon (v0.3.1)

Mikecranesync:
## Summary

v0.3.0 (#935) shipped the new hero cartoon HTML reference, but the asset itself 404s in production because Hono's `serveStatic` registrations in `src/server.ts` never covered `/images/*` — only `/public/*` and a handful of explicit per-file routes.

This is the same root cause as the pre-existing `/images/app-screenshot-desktop.png` 404 that PR #933 flagged as a known issue. Fixing both with one new route:

```ts
app.use("/images/*", serveStatic({ root: "./public" }))
```

## Verification on the live VPS once merged

- `curl -I https://factorylm.com/images/hero-fault-lookup-cartoon.png` → expect HTTP 200 with `content-type: image/png` and ~3.2 MB body
- `curl -I https://factorylm.com/images/app-screenshot-desktop.png` → expect 200 (the file exists in `public/images/`, was just unreachable)
- factorylm.com homepage renders the cartoon hero

## Test plan
- [x] All existing home.test.ts assertions pass (route addition doesn't change rendered HTML)
- [ ] Post-deploy: image GET returns 200, content-type is image/png


Mikecranesync:
Closing — re-doing as a fresh branch from main to avoid merge-conflict noise from the prior squash. Same fix lands in #937.

PR #987, opened 2026-05-04
feat(telegram): SQLite-backed durable queue for multi-photo bursts

Mikecranesync:
## Summary

Replaces the in-memory `PHOTO_BUFFER` dict in `mira-bots/telegram/bot.py` with a two-tier system: a small 4-second in-memory pre-collector that decides single vs multi, plus a durable SQLite-backed `PhotoBatchQueue` that drives a single async worker. Bot container restarts no longer drop in-flight multi-photo bursts.

Builds on PR #986 (the crash-fix that shipped `engine.process_multi_photo`).

## Problems with the old buffer

- 50-photo spam had no upper bound → could OOM the bot container.
- Vision is GPU-bound on a single Ollama node, but multi-user bursts had no fairness/serialization.
- `task.cancel()` + `create_task` on every photo had a small race window where two flush tasks could both pop the buffer.
- Bot restart silently dropped any photos still in `PHOTO_BUFFER`.

## Architecture

- **`_BURST_COLLECTOR`** (in-memory, per-chat) — same 4s sliding window as before. Holds photos awaiting the single-vs-multi decision. Lost on restart, but worst-case data loss = 4 seconds of incoming.
- After window closes:
  - **1 photo** → `_dispatch_single_photo` (existing FSM-aware `ChatDispatcher` path, unchanged behaviour while we still hold the original `Update`).
  - **2+ photos** → `_enqueue_multi_photo_burst` pushes every photo into the durable queue and sends an immediate ack.
- **Single async worker** (`_photo_batch_worker`, spawned in `_startup`) drains the queue FIFO, calls `engine.process_multi_photo`, delivers reply via `application.bot.send_message`. No `Update` object needed → works fine after restart.
- Worker concurrency hard-coded to **1** because vision is the bottleneck and parallel workers would just fight on Ollama.

## Bounds (with user-visible messages)

| Constant | Value | When hit |
|---|---|---|
| `MAX_PHOTOS_PER_BURST` | 10 | Burst capped: `"⚠️ Burst capped at N/M photos (rejected dropped)"` |
| `MAX_QUEUE_DEPTH` | 20 | `"🚧 The photo queue is full right now. Please retry in a moment."` |
| `BURST_WINDOW_SECONDS` | 4.0 | (sliding window before the collector flushes) |

## Recovery

On worker startup, `recover_orphans()`:
1. Flips any `processing` rows back to `queued` (assumes the previous worker died mid-job — vision is idempotent, so retry is safe).
2. Re-notifies the asyncio queue for every existing `queued` row so they get drained on resume.

## Schema

Single table, WAL mode (per `python-standards.md` SQLite policy):

```
photo_batches(id, chat_id, platform, status, caption, photos_json,
              photo_count, ack_message_id, created_at, queued_at,
              started_at, completed_at, error_message, reply_text)
```

DB path defaults to alongside `MIRA_DB_PATH`; override via `PHOTO_QUEUE_DB_PATH`.

`status` lifecycle: `collecting → queued → processing → done | failed`.

## Tests

`tests/test_photo_batch_queue.py` (15 tests):

- Producer: first-photo creates batch, second-photo joins, burst cap enforced, different chats don't share.
- Caption: first real caption survives default-placeholder followups (Telegram media-group caption-on-first-only quirk).
- Consumer: dequeue returns oldest first (FIFO), blocks until close, idempotent close, mark_done/mark_failed records reply/error.
- Recovery: orphan `processing` reset on worker restart, queued rows re-notified.
- Ack message ID round-trip + late-set.

All 30 photo-related tests still green:

```
tests/test_photo_batch_queue.py     ............... [15 passed]
tests/test_telegram_photo_buffer.py ........        [ 8 passed]
tests/test_multi_photo.py           .......         [ 7 passed]
```

## Test plan

- [x] Unit tests pass (30/30).
- [x] `python -m ruff check` clean.
- [x] `python -m ast.parse` confirms `bot.py` parses.
- [ ] Mike sends a 4-photo burst on production bot — gets queued ack + combined reply.
- [ ] Mike sends a single photo on production bot — still hits the FSM-aware dispatcher path (unchanged behaviour).
- [ ] Mike restarts the `mira-bot-telegram` container mid-burst — multi-photo burst resumes from the queue (orphan recovery in `_startup` logs).
- [ ] Mike sends 11 photos in one burst — gets `"⚠️ Burst capped at 10/11..."`.

## Linear

Mirror ticket to be filed and linked in a follow-up comment.

### DIFF
--- a/mira-bots/shared/photo_batch_queue.py
+++ b/mira-bots/shared/photo_batch_queue.py
@@ -1,0 +1,329 @@
+"""SQLite-backed durable photo-batch queue for multi-photo bursts.
+
+Replaces the in-memory ``PHOTO_BUFFER`` dict in ``mira-bots/telegram/bot.py``.
+Telegram users sending multiple photos in rapid succession (a "burst") get
+them collected into a single batch, queued for processing, drained by a
+single async worker (vision is GPU-bound on one Ollama node), and the
+result delivered back to the chat.
+
+Why SQLite-backed: bot container restarts no longer drop in-flight bursts.
+Why single worker: vision is the GPU bottleneck — multiple workers fight
+for the same Ollama node, so concurrency = 1.
+
+Schema (one table, ``photo_batches``):
+    id               INTEGER PRIMARY KEY
+    chat_id          TEXT
+    platform         TEXT          'telegram' | 'slack'
+    status           TEXT          'collecting' | 'queued' | 'processing'
+                                   | 'done' | 'failed'
+    caption          TEXT
+    photos_json      TEXT          JSON list of base64 vision-resized bytes
+    photo_count      INTEGER
+    ack_message_id   INTEGER       NULL until adapter sends ack
+    created_at       REAL
+    queued_at        REAL
+    started_at       REAL
+    completed_at     REAL
+    error_message    TEXT
+    reply_text       TEXT
+
+Bounds:
+    MAX_PHOTOS_PER_BURST = 10      photos past cap raise BurstFull
+    MAX_QUEUE_DEPTH      = 20      queued+processing past cap raise QueueFull
+    BURST_WINDOW_SECONDS = 4.0     wait window before flipping collecting -> queued
+"""
+
+from __future__ import annotations
+
+import asyncio
+import json
+import logging
+import sqlite3
+import time
+from dataclasses import dataclass
+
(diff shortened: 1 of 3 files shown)


Mikecranesync:
Linear mirror: https://linear.app/cranesync/issue/CRA-58

PR #1579, opened 2026-05-29
fix(hub): Channels connected counter + hide synthetic accounts from admin list (QA 2026-05-28 #3, #8)

Mikecranesync:
## What & why

WS-B of the [Hub QA fix plan](../blob/fix/hub-qa-channels-counter-admin-filter/docs/plans/2026-05-28-hub-qa-fixes.md) for the 2026-05-28 `app.factorylm.com` QA report. Fixes the **two real, fixable Hub code bugs** from that report. The rest of the report was triaged as demo-seed data (not bugs), a separate module (`mira-scan-monday`), or infra/runtime (the 500/503s) — see the plan doc.

### #3 — Channels "0 connected" vs Atlas Connected
`connectedCount` summed only the localStorage-backed `connections` map. Atlas CMMS is an always-live built-in rendered as an info-only card and never lands in that map, so the header showed "0 connected" while the card said "Connected". Now counts Atlas's built-in connection. Email stays uncounted (it's a derived info card mirroring google/microsoft — counting it would double-count).

### #8 — Test/automation accounts in the admin user list
`playwright-probe@`, `e2e-trial-*`, `NeonDB Proof`, `E2E Audit`, etc. polluted the production user list. New `isSystemAccount()` classifier in `lib/users.ts`; `/api/admin/users` filters them **by default** and exposes them via `?includeSystem=1`. The page gains a "Show N system accounts" toggle — nothing is hidden permanently, the classifier only sets default visibility and never deletes.

## Test / verification
- New unit test `system-account.test.ts` — covers the synthetic patterns and guards against false-positives on real users (e.g. `latest.news@…` is NOT flagged). Passes.
- `tsc --noEmit`: no new errors from these files (7 pre-existing errors live in unrelated `*.test.ts` files).
- `eslint`: no new errors — both pages carried 1 pre-existing `react-hooks/set-state-in-effect` each (verified by stashing); left untouched per surgical-changes principle.

## Notes
- **Screenshot rule:** these are visible Hub UI changes (Channels header count; admin "Show system accounts" toggle). Screenshots not captured here because the Hub runs on CHARLIE/VPS with a seeded DB, not locally on BRAVO. Recommend capturing during the staging-gate verification before merge.
- Scope: deliberately excludes the demo-seed "bugs" (4-vs-47 assets, dup Stardust Racers, overdue WOs) and the reproduce-first items (Review Queue 500, RSC 503s) — those are WS-A/C/D in the plan.

### DIFF
--- a/docs/plans/2026-05-28-hub-qa-fixes.md
+++ b/docs/plans/2026-05-28-hub-qa-fixes.md
@@ -1,0 +1,95 @@
+# Hub QA Fix Plan — `app.factorylm.com` (mira-hub)
+
+**Date:** 2026-05-28
+**Source:** QA Evaluation Report (Claude QA Engineer Mode), overall grade B− (74/100)
+**Author:** Claude Code (BRAVO)
+**Status:** PROPOSED — awaiting scope confirmation before execution
+
+---
+
+## 0. The one insight that reshapes everything
+
+The QA report graded by **observed severity on prod**. That is not the same as **"real, fixable Hub code bug."** Two of the three highest-severity items are almost certainly **not code defects**:
+
+- **Review Queue 500** — `src/lib/review-queue.ts::getReviewQueue` already wraps every data source (`fetchProposals`, cartoons, screenshots, audits) in individual `.catch() → []` handlers. No single source can bubble a 500. The only remaining throw paths are in `src/lib/tenant-context.ts::withTenantContext`: `pool.connect()`, `SET LOCAL ROLE factorylm_app`, or `set_config(...)`. → Likely a **prod environment gap** (`factorylm_app` role / RLS migration not applied to prod NeonDB) **or deploy lag** (prod running an older build from before the per-source `.catch()` hardening). Not a blind code edit.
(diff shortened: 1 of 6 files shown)


Mikecranesync:
### Consumer verification (post-open)

Checked existing consumers of the admin/users API before relying on CI:

- **No route unit test** for `admin/users` → the `GET()` → `GET(req: Request)` signature change has no no-arg caller to break.
- **No e2e asserts the synthetic accounts appear in the admin list.** The `e2e-trial-*` / `NeonDB Proof` / `drawer-test-*` accounts are *created* by login/persistence proof specs (`proof-pr-749-login-gate`, `proof-neondb-persistence`, `drawer-screenshot`) that assert on their own session — not on list membership. Default-hiding them leaves the DB rows and auth path intact, so those specs are unaffected.

### Residual notes (known/bounded — not reworking here)
- `tests/e2e/hub-validation.spec.ts:651` ("Users table renders with ≥5 rows") selects `tbody tr`, but the page renders card `<div>`s — it already matches 0. Pre-existing staleness, unrelated to this PR.
- The `/(^|[-_.+])test[-_.@]/i` email matcher is the loosest rule and could catch a real `qa-test@customer.com`. Acceptable: the toggle makes it recoverable and it never deletes. Unit test guards the obvious false-positive (`latest.news@`).
- Channels counter counts *distinct connections*, not connected-looking cards — the Email mirror card can show "Connected" (when Google/MS is linked) without incrementing the count, by design. Staging screenshot will surface it if anyone disagrees with that model.

PR #1582, opened 2026-05-29
fix(hub): finish login a11y (password form + toggle) + authenticated review tooling

Mikecranesync:
Addresses the 5 blind spots from the 2026-05-28 `app.factorylm.com` web-review (`/goal`). Builds on **#1581**, which landed the magic-link half of the login a11y fix while this was in flight.

### #4 / #5 — Login form (extends #1581)
#1581 fixed the **magic-link** email input (htmlFor/id, `method=post`, min-w). It left the collapsible **password** form and the toggle untouched. This completes them:
- `htmlFor`/`id` association on the password-form **email + password** inputs (were visible-text labels only — screen readers couldn't link them).
- `method="post"` on the password form (no-JS fallback reloads the login route instead of risking credentials in the URL).
- `min-h-[44px]` + `py-2` on the **14px-tall** "Sign in with password" toggle, plus `aria-expanded` (WCAG 2.5.5 target size + ARIA).
- Inline "Start free trial" link keeps the WCAG 2.5.5 inline-text exception (unchanged).

### #1 — No unauthenticated view of product health → `scripts/capture-review-session.ts`
Non-interactive login-capture that writes a Playwright `storageState` for synthetic monitors / the `web-review` skill to inject and reach authed routes. **Reuses the real credentials provider** (no JWE forging → no `AUTH_SECRET`, no NextAuth drift). **Hard prod-guard**: refuses `app.factorylm.com` unless an explicit dedicated account is supplied (prevents creating a test account on prod — cf. #1579). Complements the existing *interactive* `create-auth-state.ts`. Skill wired via `web-review/SKILL.md`.

### #2 — Soft-404 (disposition: no code change)
Both the "unknown route → 200" and "404 has no home link" findings were **artefacts of the auth wall**: the edge probe was anonymous, so middleware redirected it to `/login` before Next's 404 handler ran. `not-found.tsx` already has a home link and returns 404 for authenticated users; anon→login is **intentional** (don't leak route existence; `robots.txt` already `Disallow: /`). The real residual — broken links invisible to *unauthenticated* monitors — is closed by the #1 authenticated crawl. Full reasoning in `docs/authenticated-review.md`.

### #3 — `?_rsc=` 503s (disposition: runtime/infra, confirm on staging)
Public routes returned **200** on `?_rsc=` here, so the 503s are auth-gated/load-specific (VPS-OOM pattern), not a Next.js defect. `docs/authenticated-review.md` has a staging procedure to confirm with the #1 session.

## Verification
- `tsc --noEmit`: no new errors. `eslint` on changed files: clean.
- The #1 capture script + #3 procedure are **not yet run** — they need the staging Hub (`127.0.0.1:4101` tunnel), not reachable from BRAVO. Deliverable is script + doc + procedure, to run once staging is up. Stated plainly in the doc.
- **Screenshot rule:** visible login-UI changes; Hub runs on CHARLIE/VPS, not capturable from BRAVO — defer to staging-gate screenshots (same as #1579).

### DIFF
--- a/.claude/skills/web-review/SKILL.md
+++ b/.claude/skills/web-review/SKILL.md
@@ -161,7 +161,12 @@ If the user passes a sitemap or asks to "audit the whole site":
 
 ## What this skill does NOT do (yet)
 
-- **Auth'd routes** — no cookie injection in v1
+- **Auth'd routes** — no cookie injection built into the passes. For the
+  auth-gated Hub (`app.factorylm.com`), capture a real **staging** session first
+  via `mira-hub/scripts/capture-review-session.ts`, then open the Playwright
+  context with `{ storageState: "tests/e2e/.state/review-session.json" }` to
+  review authed routes (feed, namespace, proposals, assets, admin) instead of
+  bouncing to `/login`. See `mira-hub/docs/authenticated-review.md`.
 - **Form submission / fuzzing** — destructive; would need explicit `--aggressive` and target allowlist
 - **DAST / OWASP ZAP** — heavy, gated for v2 (`--security` flag)
 - **Persona journeys via browser-use** — gated for v2 (`--persona`)
--- a/mira-hub/docs/authenticated-review.md
+++ b/mira-hub/docs/authenticated-review.md
@@ -1,0 +1,85 @@
+# Authenticated review & monitoring of the Hub
+
+Closes three blind spots from the 2026-05-28 `app.factorylm.com` web-review.
+The Hub is auth-gated: every route 302s to `/login` for anonymous requests, so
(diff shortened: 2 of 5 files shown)


Mikecranesync:
**Tap-target precision (web-review #5):** this PR fixes the two clear sub-44px hits — the 14px 'Sign in with password' toggle (now `min-h-[44px]`) and the icon-only magic-send button (`min-w-[44px]`). Two remain, consciously left: the **password-reveal eye toggle** (42×39 — secondary control, has `aria-label`) and **'Continue with Google'** (measured 39, but nominal `h-11`=44px). Both are defensible per WCAG 2.5.5; flagging rather than silently leaving.

Issue #1662, opened 2026-06-02
Phase 3 — Route kg_writer through a proposal-transition helper (stop auto-verifying edges)


Mikecranesync:
**Context.** `mira-crawler/ingest/kg_writer.py:144` writes `kg_relationships` DIRECTLY at confidence 1.0, bypassing `relationship_proposals` — every ingested edge lands as a verified fact, violating the "never auto-verify" doctrine (.claude/CLAUDE.md). Same bypass in schematic `upsertSchematicComponents()` (confidence=1.0). The ADR-0017 helpers `proposal_transition.py` / `proposal-transition.ts` do not exist (verified-absent).

**Do.**
- Implement `mira_bots/shared/proposal_transition.py` + `mira-hub/lib/proposal-transition.ts` per ADR-0017.
- Re-route `kg_writer.upsert_relationship` → `relationship_proposals` + an `ai_suggestions(kg_edge)`; only `kg_relationships` insert on human approval.
- Audit: `INSERT INTO kg_relationships` outside the helper = 0.

**Acceptance.** Fresh manual ingest → rows in relationship_proposals, not kg_relationships. /proposals shows them; approve → verified edge. Preserve the verified-default backfill (migration 029).

Master plan: docs/plans/2026-06-01-mira-master-architecture-plan.md Phase 3. Run `codegraph_impact` on `upsert_relationship`.

Mikecranesync:
🤖 Claiming this on the **BRAVO** node — implementing in an isolated worktree on branch `feat/phase3-proposal-transition-1662`. Scope: `mira_bots/shared/proposal_transition.py` + `mira-hub/lib/proposal-transition.ts` per ADR-0017, re-route `kg_writer.upsert_relationship` through it, audit `INSERT INTO kg_relationships` outside the helper = 0. Will open a small PR. Picked per EPIC #1666 rule (unblocked ready-for-agent, off the Ignition critical-path chain the gap-closure driver is currently on).

Mikecranesync:
[...]

PR #1702, opened 2026-06-04
feat(connectors): generic connector framework + 5 mock connectors + 67 tests

Mikecranesync:
## Summary

Generic connector framework for ingesting external CMMS/SCADA/historian context into MIRA's knowledge graph, with mock connectors for the major systems and a full test suite.

**Commits (5):**
- `a264251e` connector framework base + Maximo mock connector
- `5964f37e` Ignition SCADA, SAP PM, MaintainX, AVEVA PI mock connectors
- `01d48b47` end-to-end demo + test suite (67 tests)
- `be8231cf` connector-framework.md + module README
- `dd6d1228` run mira-connectors test suite in the Python test job (CI)

## Scope / safety
- No safety-critical files touched (`engine.py`, `guardrails.py`, `uns_resolver.py`, `citation_compliance.py` all untouched).
- Connectors are mock/framework-level — they feed proposals through the existing technician-confirmation gate, not auto-verify.
- 67 tests included; CI job wired.

## Stacking
Based on `feat/hub-command-center` (the 4 shared base commits: YouTube uploader + asset-graph docs + eval-fixer + promo). Retarget to `main` once that base lands, or merge base first.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1716, opened 2026-06-04
feat(ingest): route kg_writer edges through proposal path (no auto-verify) — Phase 3 #1662 (Path-A slice)

Mikecranesync:
Closes part of #1662 (Phase 3). **Path-A vertical slice** — the live `store.py` ingest hot path. Built on the **BRAVO** node.

## Problem
`kg_writer.upsert_relationship` wrote `kg_relationships` directly at confidence 1.0 — every ingested edge landed as a **verified fact**, violating the "never auto-verify" doctrine (`.claude/CLAUDE.md`, ADR-0017).

## What this does
- **New `mira-crawler/ingest/proposal_writer.py`** — create-path proposer. An ingest edge now lands as:
  - `relationship_proposals` (`status='proposed'`)
  - `relationship_evidence` (cites the source manual chunk)
  - bridging `ai_suggestions(suggestion_type='kg_edge', status='pending')` whose `extracted_data.relationship_proposal_id` lets the Hub `/proposals` queue render it.
- **Reroute `kg_writer.upsert_relationship`** (live `store.py` path) through the proposer — it no longer touches `kg_relationships`. Signature unchanged; entity types derived from `kg_entities`, so `register_equipment_and_manual` / `register_fault_code` are untouched.
- **Relation-type vocabulary remap.** `relationship_proposals.relationship_type` is CHECK-constrained to UPPERCASE canonical; ingest emits lowercase. Mapped direction-preserving: `has_manual`→`HAS_DOCUMENT`, `has_fault`→`HAS_FAILURE_MODE`. Unmapped types are skipped with a warning (never proposed with a wrong/CHECK-violating type). On approval, `decide/route.ts` copies `relationship_type` verbatim into `kg_relationships`, so approved edges carry the canonical type.
- **RLS tenant-context.** Sets `app.current_tenant_id` via `set_config` before the inserts (mirrors `load_manifest_to_kg.py`) so they land under the RLS policies on these tables. Harmless when the ingest role bypasses RLS (owner; tables are `ENABLE`, not `FORCE`), essential under `factorylm_app`.
- **Idempotent** against both open proposals and already-verified edges → re-ingest does not flood the queue.

## ⚠️ Behavioral change (intended doctrine, flagged for coordination)
After this, the `store.py` ingest path emits **proposals, not verified edges**. The engine reads only `approval_state='verified'`, so freshly-ingested `has_manual`/`has_fault` edges become diagnosable **only after an admin approves** via `/proposals`. The approve→verify loop already exists (`mira-hub/.../proposals/[id]/decide/route.ts:115`), so nothing is stranded — but bulk OEM ingest will now generate a review queue instead of auto-densifying the graph. **If you'd prefer this behind a feature flag for the first bulk run, say so and I'll add one.**

## Verification (honest scope)
- ✅ **Mock-level (unit):** `ruff` clean; **12/12** tests in `mira-crawler/tests/test_proposal_writer.py` prove the code issues `INSERT INTO relationship_proposals` + `ai_suggestions` and **never** `kg_relationships`, plus canonical mapping, RLS tenant-context, idempotency, and self-edge/unmapped/already-verified skips. The `FakeConn` proves *which SQL is issued*, not that it executes against the live schema.
- ✅ **Static schema checks (Bravo):** confirmed `ai_suggestions.id` / `relationship_proposals.id` have `DEFAULT gen_random_uuid()`; confirmed the `relationship_type` / `evidence_type` / `source_kind` CHECK enums are satisfied; confirmed RLS policies + added tenant-context to match.
- ⏳ **Real-DB smoke PENDING:** one `propose_relationship` against dev/staging Neon + `SELECT` the rows would validate RLS + column names end-to-end. Not run from this Bravo session (prod-guard; no dev/staging Neon reachable here). **Do not merge on the unit green alone** — recommend a staging smoke first.
- ℹ️ Caller regression: kg/proposal tests pass; the broader `mira-crawler/tests/` failures are pre-existing environmental (`No module named 'mira_crawler'`/`feedparser`), none reference this code.

## Notes
- `properties` is no longer persisted at propose time (its `reasoning`/`note` keys become the proposal rationale). Path-A callers (`register_*`) pass no properties, so nothing is lost here; flagged because the end-to-end approval path also doesn't write `properties`.

## Deferred follow-ups (NOT in this slice)
The issue's global "audit = 0" spans ~10 cross-language sites; this slice scopes to the live Python path. Remaining:
- `mira-crawler/tasks/full_ingest_pipeline.py` inline inserts (cron bulk OEM loader, **Path B**).
- `mira-connectors/mira_connectors/store.py` — overlaps active PR #1702; left to avoid conflict.
- **TS side**: `mira-hub/lib/proposal-transition.ts` + 8 hub insert sites + ADR-0017 canary SQL (`tests/canary/proposal_state_drift.sql`) — needs a hub build; not Bravo-finishable.

## Placement note
Helper lives in `mira-crawler/` (not `mira_bots/shared/` as ADR-0017's text suggests) because every create-path caller is in this service/container; a cross-service runtime import would break container isolation (root `CLAUDE.md` Hard Constraints).


Mikecranesync:
## Update — review steer addressed (flag + staging smoke + audit)

**1. Feature flag added.** `MIRA_KG_INGEST_AUTOVERIFY` (default OFF). Proposals are the safe default everywhere; the legacy direct-to-`kg_relationships` auto-verify path is gated behind a deliberate opt-in for a one-time bulk migration / debug. dev/staging unset → proposals; production/bulk must explicitly set it.

**2. Staging smoke RUN and PASSED** ✅ (`mira-crawler/tools/smoke_proposal_writer.py`, against the `factorylm/stg` Neon branch under a throwaway tenant, self-cleaning):
```
proposals=1 evidence=1 suggestions=1 kg_relationships=0 rel_type='HAS_DOCUMENT'
SMOKE PASS — ingest edge proposed, not auto-verified; idempotent; cleaned up.
```
The real-DB run validated RLS tenant-context, every column name, the CHECK enums, and **zero auto-verified edges on the live path**.

**⚠️ It also caught a real bug the unit mock could not:** `relationship_proposals.created_by` is CHECK-constrained to `('llm','human','import','rule')` — my `'ingest:kg_writer'` violated it, the insert threw, and the edge was silently dropped. Fixed: `created_by='import'` (matches `load_manifest_to_kg.py`); descriptive label moved to `ai_suggestions.proposed_by='import:kg_writer'`. This is exactly why the staging smoke was the right gate.

**3. Write-path audit** added: `docs/audits/2026-06-04-ingest-edge-write-paths-1662.md` — all 14 edge-creation sites classified (proposalized / safe / gated-opt-in / still-auto-verifies / unknown), with the checklist to close #1662.

**Verification now: unit (17/17) + staging smoke PASS + static schema checks.** This addresses the "don't merge on unit-green alone" caveat — the no-auto-verify invariant is proven on the live path against staging. #1662 stays OPEN for the deferred sites (Path B, TS, canary) per the audit.

Mikecranesync:
## Update 2 — Path B done (full_ingest_pipeline reroute)

The cron bulk OEM loader (`full_ingest_pipeline.py`, `kb_growth_cron`) was the last live **Python** path still auto-verifying. It now proposes too — so the "first bulk OEM ingest" concern is closed on the Python side.

- New `propose_relationship_cursor` — psycopg2 sibling of the proposer (same invariants; uses the caller's cursor so it sees entities created uncommitted in the same transaction). Shares `_kg_edge_suggestion_fields` with the SQLAlchemy version.
- Both inline edge inserts consolidated into one gated `_write_kg_edge` (proposes by default; legacy insert only under `MIRA_KG_INGEST_AUTOVERIFY`).
- Flag centralized in `proposal_writer`, shared by both paths.

**Staging smoke extended to BOTH paths — PASS on `factorylm/stg`:**
```
[Path A] proposals=1 evidence=1 suggestions=1 kg_relationships=0 rel_type='HAS_DOCUMENT'
[Path B] proposals=1 evidence=1 suggestions=1 kg_relationships=0 rel_type='HAS_DOCUMENT'
SMOKE PASS — both ingest paths propose, not auto-verify; idempotent; cleaned up.
```

**Audit:** every `INSERT INTO kg_relationships` under `mira-crawler/` is now inside a `MIRA_KG_INGEST_AUTOVERIFY`-gated branch. Tests: 22 unit + dual-path staging smoke. The **entire Python ingest surface is proposalized**; only the TS hub writers remain (not Bravo-finishable — see `docs/audits/2026-06-04-ingest-edge-write-paths-1662.md`).

Mikecranesync:
#1716 is the **Python-side closure** for #1662 Path A/B. It makes `mira-crawler/ingest/store.py` (via `kg_writer`) and `full_ingest_pipeline.py` propose `kg_relationships` by default, gates direct inserts behind `MIRA_KG_INGEST_AUTOVERIFY`, and includes dual-path staging smoke proof on `factorylm/stg`. Remaining #1662 work is intentionally **hub-side**: TypeScript relationship-writer alignment, the `hierarchy-backfill.ts` deliberate-vs-inferred decision, the ADR-0017 canary, and a CI audit guard. **#1662 should remain open after this PR merges.** Scope is Python + the audit doc only (6 files, no TS / migrations / engine changes); CI is green (15/15), mergeable=CLEAN, no collision with the 7 newer main commits (all engine/hub, none under mira-crawler/ingest).

Issue #1596, opened 2026-05-30
bug(kb): manufacturer de-duplication — OCR/extraction variants pollute the catalog


Mikecranesync:
From the 2026-05-28 Hub QA report (Knowledge Base section; `docs/plans/2026-05-28-hub-qa-fixes.md`, WS-D).

## Symptom
The KB manufacturer catalog (Hub → Knowledge Base, ~304–323 manufacturers) contains duplicate / OCR-variant entries that should normalize to one canonical vendor, e.g.:
- `Allen-Bradley` vs `Alien-Bradley`
- `Coffing` vs `Cofemo` vs `Cofing` vs `Cottins`
- `Orlando Rigging` vs `Orldndo Rigging`
- `Deshazo` vs `Deshaco` / `Desha` / `Deshao` / `Deshazzo`

These inflate the manufacturer count and fragment chunks/manuals across spurious vendors, hurting retrieval grouping.

## Where
Extraction/normalization belongs in the ingest pipeline — `mira-crawler/ingest/` (manufacturer extraction + the UNS slug builders in `mira-crawler/ingest/uns.py`). This is a data-quality task, **not** a Hub render bug (the Hub faithfully shows what's in the KB).

## Suggested approach
- Add a manufacturer normalization/alias step (canonical-name map + fuzzy match, e.g. Levenshtein/token-sort threshold) at ingest, before UNS path construction, so variants collapse to one canonical vendor + slug.
- Backfill: a one-off reconciliation over existing `kg_entities` / KB rows to merge existing variants (dev → staging → prod).
- Respect `.claude/rules/uns-compliance.md` (path builders, slugs).

## Acceptance
New ingests map known OCR variants to the canonical vendor; the catalog no longer shows `Alien-Bradley`-class duplicates after backfill.

Mikecranesync:
Partial fix in #1713 — the **ingest-side OCR-collapse slice**. New ingests now collapse the named OCR/extraction variants (Allen-Bradley/Alien-Bradley, Coffing/Cofemo/Cofing/Cottins, Orlando Rigging/Orldndo Rigging, Deshazo/Deshaco/Desha/Deshao/Deshazzo) to the cleanest observed spelling, normalized at the `kg_writer` boundary before any UNS path is built. A logged-not-merged fuzzy proposer is included for the long tail (never auto-merges).

Two things are **deliberately carved out** of #1713 and remain open:

**1. Brand-vs-parent canonical conflict (needs a product decision).**
[...]

PR #1719, opened 2026-06-05
fix(ingest): #1596 carve-outs — AB→Rockwell decision + mira-core/mira-hub normalizers + consistency guard

Mikecranesync:
Implements the three carve-outs tracked on #1596. **Stacked on #1713** (base = `fix/manufacturer-ocr-normalize-1596`); retarget to `main` once #1713 merges.

## Carve-out (a) — brand-vs-parent decision: **Rockwell Automation (parent)**
Product decision made: Allen-Bradley canonicalizes to the corporate parent **"Rockwell Automation"**, matching the query-side resolver's `VENDOR_ALIASES` so the KB catalog and the query side group identically (ends the catalog-vs-resolver split).
- Crawler `OCR_VARIANT_ALIASES`: clean `Allen-Bradley` + OCR variants → `Rockwell Automation`.
- The resolver (`mira-bots/shared/uns_resolver.py`) is **not touched** — it already maps AB→Rockwell, so this only makes ingest agree (zero collision with active engine work).
- Long-tail rigging/hoist OEMs keep the cleanest-observed-spelling rule (resolver has no opinion → divergence-free).

## Carve-out (b) — the two other writers of `knowledge_entries.manufacturer`
The Hub KB catalog `GROUP BY knowledge_entries.manufacturer` spans every service that writes that column. #1713 covered mira-crawler; this covers the rest:
- **mira-core/mira-ingest** (photo/RAG API, Python): new `db/manufacturer_normalize.py`, wired at **both** write boundaries — `insert_knowledge_entry` and `insert_knowledge_entries_batch`.
- **mira-hub** (document-upload route, TS): new `lib/manufacturer-aliases.json` + `lib/manufacturerNormalize.ts`, wired into `api/documents/upload/route.ts`. (The production `/api/uploads/*` path hands off to the crawler, already normalized; this closes the only in-hub writer.)

## Carve-out (c) + drift guard
- **`tests/test_manufacturer_alias_consistency.py`** — asserts the three vendored maps (crawler Python, mira-core Python, mira-hub JSON) are **byte-identical**, and that any key shared with the resolver's `VENDOR_ALIASES` agrees (locks `allen-bradley → Rockwell Automation` across both surfaces). This is the integrity mechanism for vendoring the map three times (forced by per-container build contexts).
- **`tools/reconcile_manufacturers.py`** — DRY-RUN reconciliation planner for the existing polluted catalog. Classifies distinct manufacturer values as alias / fuzzy-proposed / unchanged. **No write path exists.** Optional `--from-db` is read-only (`SELECT DISTINCT` only) and **default-deny**: refuses any URL not matching a known-safe staging/local marker, so a prod URL can never connect. The actual apply remains a separate, gated dev→staging→prod step (still tracked on #1596).

## Why three copies of the map
Build contexts are per-container (`./mira-core/mira-ingest`, `./mira-hub`) and two languages (Python/TS), so a single shared runtime module is impossible. The map is vendored per service; the consistency test is what keeps the copies from drifting.

## Tests — 42 total, all green in their service contexts
| Suite | Count |
|---|---|
| `mira-crawler/tests/test_manufacturer_normalize.py` | 18 |
| `mira-core/mira-ingest/db/test_manufacturer_normalize.py` | 9 |
| `tests/test_manufacturer_alias_consistency.py` | 2 |
| `tools/test_reconcile_manufacturers.py` | 7 |
| `mira-hub` vitest `manufacturerNormalize.test.ts` | 6 |

`ruff check` clean on all Python; `tsc --noEmit` clean on touched TS; no new deps.

## Still open on #1596 (not in this PR)
- The catalog-wide **apply** backfill (gated data pass; planner here produces the plan).
- Confirm the `--from-db` safe-marker allowlist against the real staging endpoint before using that mode (default-deny means an unconfirmed marker only risks a safe false-refusal).

### DIFF
--- a/mira-core/mira-ingest/db/manufacturer_normalize.py
+++ b/mira-core/mira-ingest/db/manufacturer_normalize.py
@@ -1,0 +1,73 @@
+"""Ingest-side manufacturer name normalization (issue #1596).
+
+This module mirrors ``mira-crawler/ingest/manufacturer_normalize.py`` so the
+``mira-core/mira-ingest`` service applies the SAME OCR/variant collapse at its
+own DB write boundary (``db/neon.py`` ``insert_knowledge_entry`` /
+``insert_knowledge_entries_batch``). The Hub KB manufacturer catalog is
+computed by ``GROUP BY knowledge_entries.manufacturer``; without this, OCR /
+extraction variants ("Alien-Bradley", "Cofemo", "Orldndo Rigging",
+"Deshaco") each mint their own catalog row and fragment one real vendor.
+
+The ``OCR_VARIANT_ALIASES`` map MUST stay identical to the crawler's: a
+cross-service consistency test asserts the two maps match, so a vendor groups
+the same way no matter which service ingested the chunk. This module is the
+deterministic seed-map layer only; the crawler's fuzzy *proposer* is not
+needed here (the ingest write boundary applies the seed map, nothing more).
+"""
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+
+# Curated manufacturer-variant -> canonical map. Keys are matched
+# case-insensitively with internal whitespace collapsed (see ``normalize_
+# manufacturer``). MUST stay identical to the crawler's copy in
+# ``mira-crawler/ingest/manufacturer_normalize.py`` (guarded by a
+# cross-service consistency test). Seeded from issue #1596.
+OCR_VARIANT_ALIASES: dict[str, str] = {
+    "allen-bradley": "Rockwell Automation",
+    "allen bradley": "Rockwell Automation",
+    "alien-bradley": "Rockwell Automation",
+    "alien bradley": "Rockwell Automation",
+    "cofemo": "Coffing",
+    "cofing": "Coffing",
+    "cottins": "Coffing",
+    "orldndo rigging": "Orlando Rigging",
+    "deshaco": "Deshazo",
+    "desha": "Deshazo",
+    "deshao": "Deshazo",
+    "deshazzo": "Deshazo",
+}
+
+
+@dataclass(frozen=True)
+class NormalizedManufacturer:
+    """Result of normalizing a raw manufacturer string."""
+
+    canonical: str  # the string to store / build the catalog row from
+    method: str  # "alias" | "identity"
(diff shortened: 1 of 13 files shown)


Mikecranesync:
Two clarifications after a CI-coverage + honesty pass:

**CI actually runs these guards (verified against `.github/workflows/`):**
- `tests/test_manufacturer_alias_consistency.py` — collected by the repo-root `pytest tests/` CI job (the drift guard is real, not decorative).
- `mira-core/mira-ingest/tests/test_manufacturer_normalize.py` — **moved** here from `db/` in the last commit because CI runs `pytest mira-core/mira-ingest/tests/` and does **not** scan `db/`; it would otherwise have been orphaned.
- `mira-crawler` tests run in the crawler context (#1713).
- ⚠️ **`mira-hub` vitest is NOT run by any CI workflow today** (pre-existing — there is no vitest job in `.github/workflows/`). The 6 `manufacturerNormalize.test.ts` tests pass locally but are not gated in CI. Not introduced by this PR; flagging so it's not mistaken for enforced coverage. Wiring a hub vitest job is out of scope here (CI is under active change elsewhere).

**Honest interim state for Allen-Bradley (the parent decision has a transient cost):** for the rigging vendors, new ingests merge toward the *existing* clean spelling — immediate net improvement. But AB now writes **"Rockwell Automation," a canonical not currently in the catalog**, so until the gated backfill runs, AB is temporarily *more* fragmented (legacy `Allen-Bradley` + new `Rockwell Automation` + residual `Alien-Bradley`). The change is net-positive only **after** the backfill consolidates them. Whoever watches the catalog should expect a new "Rockwell Automation" node to appear before the backfill.

Mikecranesync:
CI-enforcement pinned precisely: the consistency test is collected by the **`test-eval-offline`** job in `ci.yml` (`pytest tests/ -v -n auto -m "not network and not slow"`; the test is unmarked → included), and the mira-core test by the **`test-unit`** job (`pytest mira-core/mira-ingest/tests/`). Both fire for this PR's changed paths (not in `paths-ignore`). Caveat inherent to stacking: `ci.yml` triggers on `pull_request: branches:[main]`, so CI will show on this PR once it's **retargeted to `main`** (after #1713 merges) — not while based on the #1713 branch.

PR #1789, opened 2026-06-08
feat(hub): KG graph-analysis layer (PageRank influence + Louvain clusters) on /knowledge/map

Mikecranesync:
## What

Adds a **graph-analysis layer** to the live `/knowledge/map` knowledge graph — **PageRank** influence (node sizing) + **Louvain** community detection (cluster coloring) + a **"Key assets" (God-Nodes)** panel — rendered through the *existing* `react-force-graph-2d` canvas.

This is the additive outcome of the **Graphify-vs-Command-Center evaluation**. Verdict: Graphify is a *code-comprehension* tool (static HTML, source-code input, vis-network) and **not** a replacement for the Command Center (a UNS tree + live-telemetry surface) or the `/knowledge/map` renderer (live, multi-tenant, mutable, evidence-backed). The one thing worth borrowing is Graphify's **analysis algorithms** — applied to the *customer* KG, not the codebase. This PR borrows the algorithms, not the tool.

**Spec:** `docs/specs/kg-graph-analysis-layer-spec.md` (spec-then-build).

## Honest result — read before merging for "insight"

I ran the actual analysis against the **demo tenant** (576 nodes / 287 edges, the densest by far) on the dev DB:

- 289 raw communities → **274 real clusters (≥2 members)**, and the **largest cluster is 3 nodes**.
- The graph is a **fragmented forest** of `equipment → manual/part` stars — **no macro-structure** (almost no `equipment ↔ equipment` edges yet), so PageRank is near-flat.
- **Today the God-Nodes panel is the genuinely useful output** (it correctly surfaces the most-documented assets). Size-by-influence / color-by-cluster are near-noise until cross-asset edges densify.

So this layer is **correct, tested, and ready for when the data is ready** — it does not manufacture insight the data doesn't contain. The lever that makes it pay off is **ingest → cross-asset proposals**, not this renderer. Every other tenant (≤14 entities) is below the analysis threshold and shows a graceful "not enough connected data yet" state.

## Changes

| File | What |
|---|---|
| `mira-hub/src/lib/knowledge-graph/analysis.ts` | Pure, unit-tested. graphology (MIT) PageRank + Louvain. `MIN_EDGES_FOR_ANALYSIS=20` degenerate guard → `available:false`. `communityCount` excludes singletons (raw count is a fragmentation artifact). |
| `.../__tests__/analysis.test.ts` | Dense fixture (real structure), **realistic-sparse** fixture (→ unavailable), robustness (self-loops/parallel/dangling). |
[...]

Issue #1663, opened 2026-06-02
Solve — Hub /proposals must render ai_suggestions, not only relationship_proposals


Mikecranesync:
**Context.** `mira-hub/src/app/api/proposals/route.ts:127` reads ONLY `relationship_proposals`. `ai_suggestions` rows of type kg_entity / component_profile / tag_mapping / uns_confirmation / namespace_move (written by `photo_ingest_worker` / nameplate flow) are **never surfaced to reviewers** — so the photo→KG flywheel writes proposals nobody can approve. Verified in docs/plans/current-state-gap-closure-plan.md §2.8.

**Do.** Extend the /proposals query + UI to render `ai_suggestions` (all 6 types) alongside relationship_proposals; wire approve/reject through the (forthcoming) proposal-transition helper. Glossary discipline per .claude/CLAUDE.md (ai_suggestions vs relationship_proposals).

**Acceptance.** A nameplate-scan proposal appears in /proposals and can be approved. No double-render of kg_edge rows.

Related: kg_writer proposal-transition issue; docs/plans/current-state-gap-closure-plan.md §2.8.

Issue #867, opened 2026-04-29
fix(engine): Remaining FSM hallucinated states — DEBUG, TEST_PREP, TRIP_CLASS_INQUIRY


Mikecranesync:
Benchmark v1.2.0 still shows new hallucinated states not in the alias dict. Add: DEBUG→DIAGNOSIS, TEST_PREP→Q2, TRIP_CLASS_INQUIRY→DIAGNOSIS, CHECKING_CONTACTOR→DIAGNOSIS, INSPECTION→Q3 to fsm.py _STATE_ALIASES.

PR #1819, opened 2026-06-08
fix(hub): sync package-lock.json (graphology + unpdf) — unblocks smoke gate + prod deploy

Mikecranesync:
## Why
CI **Smoke Test** fails at `npm ci` with:
```
npm error EUSAGE: `npm ci` can only install packages when package.json and package-lock.json are in sync.
npm error Missing: graphology@0.26.0 from lock file
```
`graphology` (added in #1789) + its sub-deps and `unpdf` are declared in `mira-hub/package.json` but were never written to `package-lock.json`. **Deploy to VPS is gated on Smoke passing**, so this drift silently blocked *every* prod deploy — including the just-merged **#867 (#1817)** and **#1663 (#1818)**, which are on `main` but never reached prod (Deploy step = skipped).

A prior sync commit (`986da445`) existed on a branch but was never merged to main.

## Fix
Regenerated `mira-hub/package-lock.json` with `npm install --package-lock-only` so `npm ci` resolves the full tree. **Lock-only** change (+146/−12); node_modules untouched.

## Verification
- Local: `npm install --package-lock-only` exits clean; `package.json ↔ package-lock.json` now in sync (graphology + unpdf present in lock).
- **Real gate:** this PR's own Smoke Test run — confirm `npm ci` passes and Deploy to VPS is no longer skipped.

## Follow-up (separate)
Repo is bun-managed (`bun.lock`) but CI gates with `npm ci` against `package-lock.json`. A bun-added dep that doesn't also update the npm lock breaks the gate. Recommend aligning CI smoke to bun (`bun install --frozen-lockfile`) or adding a lock-sync guard so this can't recur silently.

### DIFF
--- a/mira-hub/package-lock.json
+++ b/mira-hub/package-lock.json
@@ -35,6 +35,10 @@
         "clsx": "^2.1.1",
         "dexie": "^4.0.11",
         "fft-js": "^0.0.12",
+        "graphology": "^0.26.0",
+        "graphology-communities-louvain": "^2.0.2",
+        "graphology-metrics": "^2.4.0",
+        "graphology-types": "^0.24.8",
         "jose": "^5.10.0",
         "libsodium-wrappers": "^0.7.15",
         "lucide-react": "^1.8.0",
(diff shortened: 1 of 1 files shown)


Mikecranesync:
Superseded by #1855 — the bun fix. Hand-regenerating package-lock.json whack-a-moled (graphology → @swc/helpers → graphology) because local npm resolution never matches CI's. #1855 switches CI smoke to `bun install --frozen-lockfile` (bun.lock is the in-sync source of truth), which fixes the gate at the root instead of chasing the npm lock.

Issue #1761, opened 2026-06-07
bug(hub): Usage "KB Chunks" tile shows 0 — usage/route.ts still tenant-filters knowledge_entries (actual 83.5k)


Mikecranesync:
## Summary
Hub **Usage** page "KB Chunks" tile shows **0** for every real (UUID) tenant, even though `knowledge_entries` holds **83,553** chunks. `/api/usage` still filters the KB count by `tenant_id`, which is the exact bug `/api/knowledge` already diagnosed and removed.

## Evidence
```
# authed as mike@cranesync.com (tenant e88bd0e8-…)
$ curl -b cookies /api/usage   -> {"allTime":{"totalWorkOrders":4,"totalKbChunks":0}, ...}
$ curl -b cookies /api/knowledge -> manufacturers sum to 83,553 chunks (Abb 1791, …)
```
- `mira-hub/src/app/api/usage/route.ts:63-66`
  ```sql
  SELECT COUNT(*) AS total_chunks FROM knowledge_entries WHERE tenant_id = $1
  ```
- `mira-hub/src/app/api/knowledge/route.ts:9-18` (comment) explains why this returns 0:
  > "The legacy ingest pipeline tagged rows with whatever `MIRA_TENANT_ID` was set in env (typically literal `'mike'`), which does not match the per-user UUID tenantIds minted by the multi-tenant signup flow (migration 008). **Filtering by ctx.tenantId returned 0 rows of 83K+ ingested chunks.**"
- The usage route's own comment (`:59-62`) only fixed the *table* (`kb_chunks` → `knowledge_entries`), not the *tenant filter* — so the tile is still 0.

## Expected
"KB Chunks" reflects the universal OEM corpus, matching the Knowledge page (~83.5k).

## Fix (precedent already in repo)
In `mira-hub/src/app/api/usage/route.ts`, drop the `WHERE tenant_id = $1` on the `kbRows` query so it matches `knowledge_entries` global counting in `knowledge/route.ts`:
```sql
SELECT COUNT(*) AS total_chunks FROM knowledge_entries
```
(If per-tenant private-doc counting is wanted later, filter on `is_private`, not `tenant_id` — same note knowledge/route.ts gives.)

## Repro
1. Log into hub, open **Usage**.
2. Note "KB Chunks" = 0 while **Knowledge** page shows thousands.

Severity: P2 (wrong dashboard metric, undersells product on the page customers see).

PR #1592, opened 2026-05-30
feat(hub): folder = brain — UNS node-centric knowledge + subtree-grounded chat

Mikecranesync:
## folder = brain — make the namespace node the unit of knowledge AND conversation

Spec: `docs/specs/uns-node-centric-knowledge-spec.md` · ADR-0019/0020 · builds on mira-ingest-v2.

Attach a manual to a `/namespace` node → it becomes BM25-citable, addressed to that node.
Ask MIRA at a node → cited answer grounded in that node's docs **and everything beneath it**
(subtree). The founder-demo thesis ("I built the namespace, that's what makes MIRA grounded")
is now literally clickable.

### Slices
- **1 — schema** (`030_knowledge_entries_chunk_anchors.sql`): chunk anchors + `ingest_route`
  discriminator on `knowledge_entries`; `hub_uploads.kg_entity_id`. Applied to dev + staging Neon.
- **2 — node ingest** (`node-knowledge-ingest.ts`): attaching a PDF to a node extracts (unpdf) →
  chunks → real `knowledge_entries` rows (`ingest_route='v2'`, `metadata.node_id`).
- **3 — subtree retrieval + node chat** (`manual-rag.ts::retrieveNodeChunks`,
  `api/namespace/node/[id]/chat`): subtree BM25 (cascade + safety hard-stop + citations).
- **4 — UI** (`NodeChat.tsx`, namespace `ContentPanel`): "Ask MIRA" at every node + citation chips.

### Key design call (RLS)
`withTenantContext` runs as `factorylm_app` (RLS-enforced, no BYPASSRLS). `hub_uploads` has **no
RLS policy/grant**, so the read path reads the chunk→node address from
`knowledge_entries.metadata->>'node_id'` (stamped atomically at ingest), expanding the subtree via
RLS-visible `kg_entities.uns_path <@`. Equally reparent-safe. **No tenant-wide fallback** (empty
subtree → "no coverage", never whole-corpus widening).

### Verification
- ✅ **Retrieval — live on dev branch under the real `factorylm_app` role** (`scripts/verify-node-subtree-retrieval.ts`, 6/6): RLS visibility, CV-101 cites both GS10+M101, GS10 cites only its own, reparent both directions, no tenant-wide fallback.
- ✅ **Route safety/guards — unit (6/6)**: safety keyword → `X-Safety-Stop` set AND no provider fetch / no DB read; 404; 503/401/400.
- ✅ typecheck + eslint clean on touched files.
[...]

PR #1841, opened 2026-06-09
fix(hub): resolve tenant-scope drift on /api/documents — closes #1833

### LINKED ISSUE #1833: fix: Tenant-scope drift on /api/documents route
Orchestrator Lens A found: `/api/documents` uses raw `pool` (RLS-bypass) and the `knowledge_entries` query has no tenant filter, despite the route comment asserting it "stays scoped to the caller's tenant." The `cmms_equipment WHERE id = $1` lookup is also not tenant-filtered. Either add the tenant filter or update the contract comment to make the intentional-universal-corpus decision explicit and provable.

Mikecranesync:
Closes #1833.

## The problem (orchestrator Lens A)
`/api/documents` read `knowledge_entries` on the raw owner pool (BYPASSRLS) with **no tenant filter**, and resolved `cmms_equipment WHERE id = $1` **without a tenant filter** — despite a comment claiming it "stays scoped to the caller's tenant."

Two real cross-tenant exposures for any authenticated/demo user:
1. **Document leak.** Post-#1592 (folder=brain), per-tenant uploads land in `knowledge_entries` under a real UUID `tenant_id` (e.g. `/api/documents/upload`, which writes `ctx.tenantId`). A universal read surfaced **tenant A's uploaded manual to tenant B**.
2. **Asset IDOR.** `cmms_equipment WHERE id = $1` let any caller resolve **any** tenant's manufacturer/model by asset id.

## Why neither "universal" nor "pure tenant scope" is correct
`knowledge_entries` is a **hybrid corpus**:
- **Shared OEM corpus** (~83.5k chunks) is tagged with a legacy non-UUID slug (`'mike'`) and is intentionally universal. Filtering it by `tenant_id = $caller` returns ~0 rows — the exact bug **#1761** fixed.
- **Per-tenant uploads** are tagged with a real UUID `tenant_id` and must be private.

So the fix is the `is_private` split that #1761's own note foreshadowed ("switch to `is_private`, not `tenant_id`, when per-tenant docs exist") — which arrived with #1592.

## The law (codified)
New rule: `.claude/rules/knowledge-entries-tenant-scoping.md`.
- **Read** (per-tenant document/RAG surfaces): `WHERE (is_private = false OR tenant_id = $caller)`, on the **raw pool** — `withTenantContext`'s RLS policy is pure `tenant_id = app.tenant_id` and would hide the OEM corpus, so it *cannot* express the hybrid.
- **Aggregate OEM surfaces** (`/api/knowledge`, `/api/usage`, library rollups) stay universal.
- **Write** (per-tenant uploads): set `is_private = true`.

## Changes
- `/api/documents`: hybrid read filter + `cmms_equipment … AND tenant_id = $2`.
- `/api/documents/upload`: writes `is_private = true`.
- `db/migrations/045_…`: backfills existing UUID-tenant rows to `is_private = true` (idempotent, reversible, gated dev→staging→prod).
- The rule doc (the law) + a rollout checklist.
- Vitest (`…/documents/__tests__/route.test.ts`, 4 tests, green): asserts the hybrid filter is emitted scoped to `$caller` and the cmms lookup carries `tenant_id = $2`.

## Verification
- `npx vitest run src/app/api/documents` → 4/4 pass.
- `npx tsc --noEmit` → my files clean; error count unchanged vs baseline (9 pre-existing in unrelated test/optional-dep files).

## Overlap with #1837 (please read before merging either)
PR #1837 (P0-2) scopes the **same** `cmms_equipment` lookup with the identical `AND tenant_id = $2` (good cross-check) but **keeps `knowledge_entries` universal**, citing the #1761 precedent. That precedent is about *aggregate* OEM views and predates the per-tenant-upload reality (#1592) — so "keep universal" still leaks uploads. **This PR supersedes that half.** Suggested: #1837 keeps its rate-limiting + cmms fix; this PR owns the `knowledge_entries` decision. They touch the same file and will conflict — sequence them.

## Tracked follow-up (rollout, not in this PR)
The law governs ~13 other `knowledge_entries` read surfaces and the production write path. This PR is the **reference implementation** for the #1833 surface; remaining work (hybrid filter on `/api/assets/[id]/documents` + RAG reads; `is_private = true` in the `mira-crawler`/folder=brain ingest write path, distinguishing customer-upload from OEM crawl) is listed in the rule doc's rollout section. Migration 045 protects all rows existing at apply time; `/api/documents/upload` is already correct.

- [ ] Run `/security-review` on the diff before merge (per #1837's gate).


Mikecranesync:
Heads-up from the beta-readiness track (#1837/#1845):

1. **Migration collision blocking merge:** this PR adds `mira-hub/db/migrations/045_knowledge_entries_private_uploads.sql`, but `045_knowledge_entries_chunk_anchors.sql` already merged to main — so **Migration Order Check will fail** until this is renumbered (048 is now taken by #1845's ledger; next free is **049**).

2. **IDOR half carried in #1837:** because this PR can't merge as-is and the cross-tenant `cmms_equipment` leak is stranger-facing, I added the identical `AND tenant_id = $2` hunk to #1837 (green, CLEAN). When this rebases, that hunk no-ops — **this PR's real contribution is the hybrid-corpus read-filter** (`is_private = false OR tenant_id = $caller`) + migration + upload route, which #1837 does **not** touch. Both are still needed.

Issue #1806, opened 2026-06-08
Blind upload doors (/api/uploads, folder, local) still write OW-KB-only — not v2 knowledge_entries


Mikecranesync:
## Context

PR #1592 (`folder = brain`, merged `6758e7e6`) closed the upload→retrieval gap **only on the deliberate node-attach door**: `POST /api/namespace/node/[id]/files` → `ingestPdfToNode` (`mira-hub/src/lib/node-knowledge-ingest.ts`) chunks a PDF into `knowledge_entries` (`ingest_route='v2'`, generated `content_tsv`, `metadata.node_id`), which `retrieveNodeChunks` + NodeChat then cite.

The **blind production upload doors still route to the Open WebUI KB only** and never reach `knowledge_entries`, so files uploaded through them are **not citable in chat**:

- `mira-hub/src/app/api/uploads/route.ts` → `mira-ingest-client` → `/ingest/document-kb` → OW KB
- `mira-hub/src/app/api/uploads/folder/route.ts` (the MiraDrop desktop-watcher door) → OW KB
- `mira-hub/src/app/api/uploads/local/route.ts` → OW KB
- `mira-hub/src/app/api/documents/upload/route.ts` → single unchunked `knowledge_entries` demo shim (not retrievable)

## Decision needed (do NOT just wire it blindly)

The research doc warns against **two divergent ingest writers** into `knowledge_entries` (this + `mira-ingest-v2`/MiraDrop ADR-0019). Pick one:

1. **Wire the blind doors to v2 ingest** — but they have no node/UNS address. `node-knowledge-ingest` requires a `kgEntityId`. A blind upload would need a target node (resolve from filename/UNS, or land in an "Inbox" node) — that's a design choice, not a 1-line wire.
2. **Deprecate the blind doors** in favor of the node-attach affordance (folder=brain is node-addressed by design), and migrate MiraDrop to attach-to-node.

## Why this is NOT a beta blocker

The beta gate's surface is the **node-attach + NodeChat** flow, which is complete and UI-reachable (empty tenant → "New Folder" → attach manual → ask). This issue is about the *other* upload entry points.

## Refs
- `docs/research/2026-06-07-upload-retrieval-gap-and-beta-path.md` §4 step 4 + post-merge banner
- `tests/beta/README.md` (post-#1592 status)
- ADR-0019 (mira-ingest-v2), ADR-0020 (knowledge node addressing)

[...]

Issue #1749, opened 2026-06-06
ops: prod VPS memory-starved + docling-serve OOM-looping (transient 502 source)


Mikecranesync:
Surfaced while diagnosing the 2026-06-06 app.factorylm.com 502 (see PR #1748, which fixes the *healer* resilience but not the underlying cause).

## Evidence (2026-06-06, prod `factorylm-prod`)
- `free -h`: **123Mi free**, **swap 3.7Gi/4.0Gi used (93%)** on an 8GB box.
- `dmesg`: `docling-serve` OOM-killed by its memcg at **09:01** and **10:01 UTC** (`total-vm:99GB`, anon-rss ~3.1GB each time). It is the only container hitting its cgroup limit — the cap is working, but it loops.
- Heartbeat caught 5 core containers "DOWN" at **02:45** — this was a **deploy** mid-recreate (docker events show exit 137 + `hasBeenManuallyStopped=true` at 02:40–02:44, deploys ran 02:44 + 02:49). Self-resolved.
- Mike's morning 502 = the **manual `workflow_dispatch` deploy at 10:44** + follow-on 10:48 recreate window. Transient, not a crash.

## Two distinct problems
1. **Memory pressure / docling OOM loop.** docling-serve repeatedly balloons to ~3GB RSS and gets killed. On an 8GB host with swap already exhausted, this destabilizes neighbors. Options: lower docling concurrency/worker memory, harder mem cap + restart backoff, move docling off the hot box, or right-size the droplet. (Prior context: the May-2026 VPS OOM incidents — PRs #1318/#1319/#1336.)
2. **Deploy-induced transient 502.** `compose down/up` recreate briefly 502s the edge. Consider rolling/`--no-recreate` where possible, or a maintenance-page/health-gate so user-facing deploys don't flash a gateway error. Also: a **manual** deploy was dispatched at 10:44 — worth confirming who/why.

## Not in scope of PR #1748
That PR makes the healer able to recreate a removed container and restores Telegram alerting. This issue is the root cause those guardrails protect against.

PR #1860, opened 2026-06-10
test(beta): gate harness speaks session-cookie auth (BETA_GATE_COOKIE) — #1806 deferred

Mikecranesync:
## What

Item **D** of the beta-gate hardening. Closes the **harness auth gap** that blocked the beta release gate from ever reaching the surface it's meant to test.

## The gap

`tests/beta/_gate.py` drives the Hub NodeChat doors (`/api/namespace/node/<id>/{files,chat}`) — the surface PR #1592 grounds (upload → `knowledge_entries` → cited answer). But those routes authenticate via a **next-auth session cookie** (`sessionOr401`), **not** the `BETA_GATE_API_KEY` bearer the harness sent. So the gate could never authenticate against the real doors. The README already flagged this as an unmet provisioning requirement.

## Fix (HTTP-only — honesty contract preserved)

- `_gate.py`: `GateConfig.cookie` + `BETA_GATE_COOKIE` env; `_headers()` forwards it as a `Cookie` header. Bearer still works for JSON engine/pipeline surfaces. The harness still **only goes through the real doors** — it never seeds `knowledge_entries`.
- `test_gate_harness.py`: 3 new `MockTransport` tests pin that the cookie header is built and sent to NodeChat (no live env).
- `README.md`: run recipe now uses `BETA_GATE_COOKIE`, documents how to mint it (the `folder-brain-proof` register→csrf→credentials flow) and the `tenants`-table mirror the provisioner must do for the chunk-insert FK.

## ⚠️ xfail deliberately NOT removed — this is the human-gated part

The goal asked to remove the xfail. After investigation that is **blocked on durable provisioning**, and removing it now would be actively harmful:

- Removing a **release-gate** xfail must leave a result anyone can re-run. A one-shot local `next dev` green **vanishes** when the hand-stood server is torn down → `pytest .../beta_ready_upload_retrieval_citation.py` then errors (`GateUnavailable`) **RED for everyone**, next session.
- So the gate needs a **durable** surface: a standing dev/staging NodeChat endpoint + tenant + session cookie + reachable LLM cascade, **or** this job wired into CI with those secrets. That is the **`#1749` / staging-provisioning** human step the goal lets me stop on.
- `ingestPdfToNode` is inline (`unpdf`, no external docling) so the flow is *locally* runnable for evidence, but local ≠ durable.
- The `test_upload_retrieval_citation.py` twin's xfail is **also** left intact — it IS collected by default pytest, so flipping it would turn CI red for everyone.

The marker's `reason` is updated to state exactly what remains and points to the repro recipe.

`#1806` (blind upload doors still write OW-KB-only) is **deferred** per the goal.

## ⚠️ Stacked on #1807 (base = `lane/upload-retrieval-gate`)

The NodeChat SSE harness this builds on (`_parse_sse_answer`, commit 586b938e) is lane-only (not on `main`). Merges after #1807.

## Verify

```
pytest tests/beta/test_gate_harness.py tests/beta/test_upload_retrieval_citation.py
# → 11 passed, 1 xfailed
```

### DIFF
--- a/tests/beta/README.md
+++ b/tests/beta/README.md
@@ -32,9 +32,20 @@ pytest tests/beta/test_gate_harness.py -v
 BETA_GATE_UPLOAD_URL=https://<staging>/api/namespace/node/<id>/files \
 BETA_GATE_CHAT_URL=https://<staging>/api/namespace/node/<id>/chat \
 BETA_GATE_TENANT=<tenant-uuid> \
-BETA_GATE_API_KEY=<token> \
+BETA_GATE_COOKIE='next-auth.session-token=<jwe>' \
 pytest tests/beta/beta_ready_upload_retrieval_citation.py -v
 ```
+
+`BETA_GATE_COOKIE` (added 2026-06-09) is the auth the Hub NodeChat routes
+actually require — they gate on a next-auth session cookie (`sessionOr401`), not
+the `BETA_GATE_API_KEY` bearer. Mint the cookie the same way
+`mira-hub/tests/e2e/folder-brain-proof.spec.ts` does: `POST /api/auth/register`
+→ `GET /api/auth/csrf` → `POST /api/auth/callback/credentials` and read the
+`next-auth.session-token` from the `Set-Cookie` response. (The bearer env still
+works for JSON engine/pipeline surfaces.) The provisioner must also mirror the
+auth-side tenant id into the data-side `tenants` table — see the
+`folder-brain-proof` `seedFixture` note — so the upload's chunk INSERTs satisfy
+the `knowledge_entries.tenant_id` FK.
 
 The harness speaks both the NodeChat contract (`messages` body + SSE response) and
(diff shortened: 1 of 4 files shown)


Mikecranesync:
Two notes:

1. **CI:** base is `lane/upload-retrieval-gate`; `ci.yml` triggers only on PRs to `main`, so no ci.yml checks appear here by design. Harness tests verified locally (11 passed, 1 xfailed) and will run in CI once retargeted to `main` after #1807.
2. **Live-run artifacts:** the evidence run left a few orphan rows on the **dev** Neon branch (a registered user + tenant `19983940-…` + one empty namespace node). Dev is safe-to-break and no `knowledge_entries` chunks were written (the upload 500'd at the grant — see #1861), but flagging it since `folder-brain-proof.spec.ts` normally cleans up in `afterAll` and this manual run did not.

PR #1861, opened 2026-06-10
fix(hub): GRANT INSERT on knowledge_entries to factorylm_app — unblocks folder=brain upload→retrieval

Mikecranesync:
## The bug (found by running the beta gate live)

PR #1592 (`folder = brain`) made an uploaded manual citable by inserting its chunks into `knowledge_entries` via `lib/node-knowledge-ingest.ts` — **under `withTenantContext`, i.e. the limited `factorylm_app` role**. But `011_grant_app_kb_access.sql` granted that role only **`SELECT`** on `knowledge_entries`, never `INSERT`.

So the real upload door **500s at the chunk INSERT**:

```
[api/namespace/node/:id/files POST] error: permission denied for table knowledge_entries
    at ingestPdfToNode (src/lib/node-knowledge-ingest.ts:103)
  routine: 'aclcheck_error'
```

**No uploaded manual ever becomes citable under the app role** — which is exactly the upload→retrieval gap the beta release gate (`tests/beta/`) exists to catch. The retrieval (SELECT) half works; the write (INSERT) half was never granted.

## How it was found

Driving the beta gate against a local `next dev` on the **dev** Neon branch with a real minted next-auth session: the flow got through real auth → the real `POST /api/namespace/node/<id>/files/` door → `unpdf` chunking → and then failed at the INSERT with `permission denied`. Confirmed against the migrations: grep shows `011` grants only `SELECT ON knowledge_entries TO factorylm_app`; no migration ever grants `INSERT`.

## The fix

```sql
GRANT INSERT ON knowledge_entries TO factorylm_app;
```

- Tenant isolation on insert is **preserved**: the RLS policy from `011` (`knowledge_entries_tenant`, `AS PERMISSIVE FOR ALL`) has no explicit `WITH CHECK`, so Postgres reuses its `USING` expression (`tenant_id = app.tenant_id`) as the INSERT `WITH CHECK`. A row can only be inserted for the caller's own tenant.
- No sequence/identity grant needed — the chunk id is `randomUUID()` app-side and `content_tsv` is a `GENERATED` column.
- `node db/check-migration-order.mjs` → **passes**.

## Apply path (gated — not hand-applied)

Apply via `apply-migrations.yml` dev→staging→prod (`dry-run` then `apply`), per `docs/environments.md`. **This is the migration half of the beta-gate unblock; once it lands on a durable surface, the beta gate's write path works end-to-end.** Pairs with #1860 (gate harness cookie auth).


Mikecranesync:
**Evidence scope (honesty note):** I proved that the missing `INSERT` grant is *the* blocker — the real door + auth + `unpdf` chunking all worked, and the run failed precisely at the `knowledge_entries` INSERT with `permission denied`; grep confirms no migration ever grants INSERT. I did **not** observe one continuous green upload→ask→cited run (the ad-hoc dev grant needed to demonstrate that was correctly blocked as an unsanctioned RBAC change). So: this **removes the identified INSERT blocker**; a full end-to-end green is to be **confirmed on the provisioned surface** after `apply-migrations.yml` lands the grant. Not an over-claim of a verified green flow.

Mikecranesync:
## 🔴 CONFIRMED: this is a live PRODUCTION bug

Read prod's grants via the sanctioned read-only `db-inspect.yml` (target=prod; the grant-drift check added to this PR). Result for `factorylm_app` on `knowledge_entries`:

| env | grant | folder=brain manual upload |
|---|---|---|
| **prod** | **`SELECT` only** | **500s at the chunk INSERT — broken now** |
| staging | `INSERT,SELECT` (hand-granted, not from a migration) | works |
| dev | `SELECT` only | 500s (reproduced in the live run) |

Every *other* write-path table (`hub_uploads`, `kg_entities`, `namespace_direct_uploads`, `relationship_proposals`, `installed_component_instances`) has full `DELETE,INSERT,SELECT,UPDATE` in prod — **only `knowledge_entries` lacks INSERT.**

**Impact:** any user who attaches a manual to a `/namespace` node on app.factorylm.com gets a 500; the manual is never chunked into `knowledge_entries` and never becomes citable. The retrieval (SELECT) half works; the write (INSERT) half has been silently failing in prod.

**Why it drifted:** the INSERT grant was hand-applied to *staging* but never encoded as a migration, so prod (and any env rebuilt from migrations, like dev) never got it. This PR's migration 049 fixes that permanently. **Recommend prioritizing the merge + `apply-migrations.yml` apply to prod.**

PR #1867, opened 2026-06-10
fix(ci): apply-migrations records migration_name, not the non-existent filename column

Mikecranesync:
## Bug

`apply-migrations.yml` records applied files with `INSERT INTO schema_migrations (filename)`, but **migration 048** (`schema_migrations_ledger.sql`) made the canonical ledger column **`migration_name`** (`UNIQUE(migration_name)`), and the workflow's own 'all'-mode skip logic already reads `SELECT migration_name`. So every `mode=apply` run commits the migration SQL and then **dies at the ledger step**:

```
ERROR: column "filename" of relation "schema_migrations" does not exist
```

Surfaced while applying migration **049** (the `knowledge_entries` INSERT grant, #1861) to staging: the GRANT committed, the ledger INSERT failed, the run went red.

## Fix

Three stale `filename` column refs → `migration_name` (seed-ledger `CREATE` + `INSERT`, and the apply-loop `INSERT`), matching migration 048 and the existing read path. The recorded value is unchanged (still the migration filename). **Unblocks every future migration apply, not just 049.**

Validated by re-running `apply-migrations.yml` (target=staging, migrations=049, mode=apply) from this branch ref — see run linked in comments.


Mikecranesync:
Validated end-to-end from this branch ref:
- **staging** apply (049, mode=apply) → ✅ green ([run 27263362172](https://github.com/Mikecranesync/MIRA/actions/runs/27263362172)): GRANT + ledger record both succeeded (previously died at the `filename` column).
- **prod** apply → ✅ green ([run 27263431990](https://github.com/Mikecranesync/MIRA/actions/runs/27263431990)).
- prod `db-inspect` confirms `knowledge_entries` now grants **`INSERT,SELECT`** (was `SELECT` only).

**Ready for your review/merge.** I did not admin-merge (branch-protection bypass on shared CI tooling wasn't in scope). The prod fix is already live and does not depend on this PR merging — this just makes the corrected ledger logic permanent on main for the next migration apply.

PR #1871, opened 2026-06-10
fix(quickstart): GTM hardening — citation relevance + e2e smoke gate (RED #2/#3)

Mikecranesync:
## Go-to-market hardening — Quickstart chat path

The #1 GTM priority: harden the stranger-facing **public Quickstart "ask MIRA"** experience on factorylm.com (`/api/quickstart/ask`). This PR closes all three RED items and adds the tracking checklist for all four GTM priorities.

📋 Checklist: `docs/gtm/go-to-market-hardening-checklist.md`

### ⚠️ Implementation note (read first)
The RED #2/#3 instructions pointed at the **Python** engine (`mira-bots/shared/engine.py` + `citation_compliance.py`). I verified via CodeGraph that **the quickstart path never touches that engine** — it's entirely TypeScript in `mira-hub` (`quickstart/ask/route.ts` → `manual-rag.ts` → cascade). A Python-side filter would not have hardened the quickstart surface at all. So the fix lives in TS, where the bug actually is. The Python bot engine (Telegram/Slack) has the *same* wrong-vendor citation gap on a *separate* surface — flagged as follow-up in the checklist, **not** fixed inline (surgical scope).

### RED #1 — Rate limit ✅ (verify-only)
PR #1838 (`fix/quickstart-rate-limit`, merge `5fd70dae`) is **already merged and on `origin/main`** — 20 req/IP-min → 429. Nothing to do; verified.

### RED #2 — Citation relevance ✅
`retrieveManualChunks` falls back to a **tenant-wide** BM25 query when the manufacturer-scoped query returns 0 hits, so a Danfoss question could cite a Siemens manual → looks like a hallucination even when the answer text is right.

- New `filterCitationsByRelevance()` in `manual-rag.ts` — lightweight string match (manufacturer / model / source-url), **no LLM call**, no `uns_resolver` (anonymous surface, no `uns_context`).
- Keeps blank-manufacturer chunks (generic/educational carve-out); drops set-but-mismatched vendors.
- Wired into `route.ts` **before `buildGroundedContext`** so the model never grounds on the dropped chunk and `[n]` markers stay aligned with the returned citations. Empties → cite-or-refuse falls through to a refusal (the correct outcome).
- 6 pure-function vitest cases incl. the Siemens/Danfoss regression. ✅ pass, eslint clean.

### RED #3 — Quickstart smoke gate ✅
- New `tests/smoke/test_quickstart_e2e.py` — **stdlib-only** (no requests/Playwright), dual-target (localhost + prod). Asserts: 200 + non-empty answer; **≥1 citation OR explicit no-docs refusal**; flood → 429 (localhost/flag-gated so prod isn't hammered). Follows 308 trailing-slash redirects with the POST body intact.
  - **Verified PASS against prod**: 394-char answer, 5 citations.
- Wired into `smoke-test.yml` as a Python step (ubuntu has `python3`, no pip).
- Also strengthened the existing Playwright deploy gate (`money-path.spec.ts`) to enforce **citation-or-refusal** — so the always-on pre-deploy gate validates the RED #2 filter end-to-end (previously only asserted `citations` was an array).

### Blast radius
`codegraph_impact`/`callers`: `retrieveManualChunks`/`buildGroundedContext` are called by `quickstart/ask` and `assets/[id]/chat`. The new `filterCitationsByRelevance()` is a standalone export with **no existing-signature changes**, wired into the single quickstart call site only.

### Verification
- `npx vitest run src/lib/manual-rag.test.ts` → 6/6 pass
- `eslint` on changed TS → clean
- `ruff check` on the Python smoke → clean; `py_compile` ok
- `python3 tests/smoke/test_quickstart_e2e.py` against `app.factorylm.com` → **PASS** (429 flood not run against prod — denied/antisocial by design; covered by the gated Playwright check)

### DIFF
--- a/.github/workflows/smoke-test.yml
+++ b/.github/workflows/smoke-test.yml
@@ -63,6 +63,17 @@ jobs:
           HUB_URL: https://app.factorylm.com
         run: npx playwright test --config playwright.smoke.config.ts
 
+      # Quickstart money-path smoke (GTM RED #3) — stdlib-only Python, no pip
+      # install. Belt-and-suspenders to the Playwright gate above: asserts
+      # /api/quickstart/ask returns a grounded answer OR an explicit refusal.
+      # The 429 flood is self-gated (SMOKE_RATE_LIMIT_CHECK) so we never hammer
+      # prod on every PR. Runnable by hand against localhost or prod; see the
+      # module docstring.
+      - name: Quickstart E2E smoke (Python, stdlib)
+        env:
+          QUICKSTART_SMOKE_URL: https://app.factorylm.com
+        run: python3 tests/smoke/test_quickstart_e2e.py
+
       - name: Upload failure screenshots
         if: failure()
         uses: actions/upload-artifact@v7
--- a/docs/gtm/go-to-market-hardening-checklist.md
+++ b/docs/gtm/go-to-market-hardening-checklist.md
@@ -33,6 +33,12 @@ BM25 chunks.
   **chunks** (before `buildGroundedContext`), so the LLM never sees the
   irrelevant content and `[n]` markers stay consistent. → `filterCitationsByRelevance()`
   in `manual-rag.ts`, wired into `quickstart/ask/route.ts`.
+  **Scope:** fires off the **picked manufacturer** (the request's `manufacturer`
+  field — exactly the fallback case where the bug bites). A stranger who picks
+  no manufacturer and types a free-text vendor mention ("my Danfoss VFD…") gets
+  `manufacturer=null` → passthrough; BM25 query-text ranking is the only defense
+  there. Free-text vendor extraction is out of scope on this surface. **Test the
+  Danfoss case via the manufacturer picker.**
 - [x] **RED #3 — Quickstart smoke gate.** End-to-end check on every deploy:
   POST a real maintenance question, assert non-empty answer + citations-or-
   explicit-refusal + no error status; assert 429 on the 21st rapid request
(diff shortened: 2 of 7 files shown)


Mikecranesync:
**Scope note (RED #2):** the citation-relevance filter fires off the **picked manufacturer** (the request's `manufacturer` field) — the exact case the tenant-wide BM25 fallback breaks. A no-picker free-text vendor mention ("my Danfoss VFD…") yields `manufacturer=null` → passthrough (BM25 query-text ranking only). Free-text vendor extraction is out of scope on this anonymous surface (no `uns_resolver`). **Verify the Danfoss case via the manufacturer picker.**

Mikecranesync:
## Wave 2 verdict: CLOSE — superseded (partly), with salvage

Assessed against current `main` after the Wave-1 hardening merges. The headline feature is **already on main, better-factored**:

- #1871 adds `filterCitationsByRelevance(chunks, manufacturer)` **inline in `manual-rag.ts`**.
- `main` already merged `stripConflictingVendors(chunks, manufacturer ?? question)`, extracted into `mira-hub/src/lib/vendor-relevance.ts` (+ `__tests__/vendor-relevance.test.ts`).

main's version is a **strict superset**: it also infers the vendor from the question text when no manufacturer is set. Resolving the `quickstart/ask/route.ts` conflict in favor of this PR would **regress** that question-fallback. So the citation-relevance code here is dead, and `manual-rag.test.ts` (which tests the dead fn) dies with it.

### Salvage (independent of the dead fn — worth a fresh small PR)
- `tests/smoke/test_quickstart_e2e.py` (+111) — black-box smoke of the stranger-facing `/api/quickstart/ask` contract (200 + cite-or-refuse + 429 rate-limit). **Directly beta-gate-aligned.**
- `docs/gtm/go-to-market-hardening-checklist.md` (+98) — standalone doc.

**Recommendation:** close this PR; cherry-pick those two files into a fresh `test(smoke): quickstart e2e + GTM checklist` PR (no code, no conflict). Queued for Mike's go in the Wave-2 go-list. Reopen if the question-fallback in main's version is judged insufficient.

PR #1874, opened 2026-06-11
fix(hub): asset-agent tenant_id is TEXT not UUID — fixes "Insert failed" on Train & approve

Mikecranesync:
## Summary
Fixes the red **"Insert failed"** on the onboarding **Train & approve before deploy** step (and the asset Validate tab) for any **slug-tenant** (e.g. the `mike` demo tenant).

**Root cause:** MIRA has a dual `tenant_id` space — the CMMS/equipment family (`cmms_equipment`, `tenants`, …) is **TEXT**-tenanted with live slug tenants like `'mike'`, while the kg/Hub family is **UUID**. Migrations `046/047` (asset-agent validation) key on `cmms_equipment.id` but declared `tenant_id UUID` (and RLS casting `app.tenant_id::UUID`). The `validation-qa` POST binds the session's TEXT tenant `'mike'` into the UUID column:

```
ERROR: invalid input syntax for type uuid: "mike"   →  UI: "Insert failed"
```

`GET …/agent-status` works (badge renders) because a brand-new asset matches **zero** rows, so the RLS UUID-cast never fires — the **first INSERT** is where the text tenant meets the UUID column and throws.

## Proof (staging, real route SQL under `factorylm_app` + tenant context)
| Tenant | Route INSERTs |
|--------|---------------|
| `'mike'` (TEXT, demo) | **FAIL** — `invalid input syntax for type uuid: "mike"` |
| `78917b56…` (UUID) | **SUCCEED** |

A clean-room repro passes *only* if seeded with a UUID tenant — which masks the defect. Verified on ephemeral pg16: pre-fix the `mike` INSERT fails; post-`048` it succeeds and UUID tenants still work; `048` is idempotent (applied twice, no error).

## Changes
- **`048_asset_agent_tenant_text.sql`** — `ALTER` both `tenant_id` columns to `TEXT`, recreate RLS policies as plain TEXT compares (matching canonical `008`). Drops policy + GiST index before the type change (policy dependency / type-specific opclass), recreates after. Idempotent.
- **`docs/tech-debt/2026-06-10-train-approve-insert-failed-diagnosis.md`** — full diagnosis + staging reproduction.
- **`.claude/rules/mira-hub-migrations.md`** (+ CLAUDE.md wiring) — codifies the lessons so the class can't recur: tenant_id type per family, in-type RLS (no cross-cast), `GRANT … TO factorylm_app`, `ALTER COLUMN TYPE` ordering, and "verify with a real slug tenant, not a synthetic UUID." GREEN-tested with a subagent.

## Promotion
- `migration-verify.yml` auto-applies `048` to **staging** on this PR.
- **Prod** needs the gated `apply-migrations.yml` dispatch after merge.

## Image #3 (folder chat "no manual coverage") — not a bug
Separate symptom in the report: the Conveyor Lab folder has 0 documents, so the grounded chat correctly refuses. Fix = attach a manual. No code change.


Mikecranesync:
⚠️ **Migration-number collision** — this PR adds `mira-hub/db/migrations/048_asset_agent_tenant_text.sql`, but `048_schema_migrations_ledger.sql` already holds **048** on `main`. The current migration head on origin/main is **`049_grant_app_knowledge_entries_insert.sql`**.

**Safe fix before merge:** renumber this file to **`050_asset_agent_tenant_text.sql`** (next free number). The `schema_migrations` ledger keys on the full filename, so apply-order tolerates the rename; just regenerate any ledger reference if one exists.

(Flagged during the GTM hardening audit — PR #1875 / #1877.)

PR #1875, opened 2026-06-11
docs(gtm): master go-to-market hardening checklist

Mikecranesync:
## What

Adds **`docs/gtm/go-to-market-hardening-checklist.md`** — the master document tracking everything needed before a stranger can use *and pay for* each MIRA surface. Points `wiki/hot.md` at it.

## Method (evidence discipline)

- Audited against **`origin/main` @`7d3483cf`** (deploy truth) via a detached worktree — **not** the local branch (which was 66 commits behind; reading it would repeat the documented false-RED failure mode in BETA_READINESS).
- Every Status grounded in a PR #, commit, `file:line`, `gh` output, or **live HTTP probe** of `factorylm.com` / `app.factorylm.com`.
- Runtime/ops facts (deploy state, billing, SSL, bot liveness) are marked `UNKNOWN` + probe-command or `HUMAN` — never code-inferred "DONE".

## Coverage

9 surfaces: public website · quickstart chat · Hub · Slack bot · Telegram bot · engine · SimLab · infrastructure · revenue. Each item carries **Status / Owner / Priority / Evidence**.

## Headline

**3 P0 blockers to the first dollar — all HUMAN:**
1. **Stripe in TEST mode** (#1831) — code is live-ready, flip Doppler `STRIPE_SECRET_KEY`→`sk_live_`.
2. **DigitalOcean VPS billing** — confirm card on file.
3. **NeonDB billing** — confirm card on file.

Engine / Hub / quickstart are 🟢 on deploy truth (rate-limit #1838, vendor-strip #1858, BM25 bound #1859 all merged; safety "hot work" + citation-enforce live). Gemini 403 (#1830) is a **degraded cascade tail**, not an outage — Groq+Cerebras carry it. SimLab fully merged (#1816/#1842/#1849/#1741).

## Note on prompt PR numbers

- Rate limit "#1838" → **MERGED** ✅
- Citation filter "#1871" → still **OPEN**, but its goal landed via **#1858 (merged)**; #1871 should be confirmed + closed as superseded.

## Files

- `docs/gtm/go-to-market-hardening-checklist.md` (new)
- `wiki/hot.md` (pointer)

Pairs with `wiki/orchestrator/BETA_READINESS.md` and `docs/hardening/hub-hardening-backlog.md`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1877, opened 2026-06-11
fix(quickstart): fault-code-aware retrieval — stop false-refusal on PowerFlex F004 demo

Mikecranesync:
## The bug (PR #1875, Surface 2)

Public quickstart chat refused the **canonical demo question** — *"My Allen-Bradley PowerFlex 525 is showing fault F004, what does it mean?"* — with *"I don't have manuals for that in the public knowledge base"* **while rendering 6 PowerFlex 525 citations.** Money-path contradiction.

## Diagnosis — the PR #1875 hypothesis was wrong (proven by live probes)

| Probe | Result |
|---|---|
| `{"question":"My Allen-Bradley PowerFlex 525 … fault F004 …"}` (verbose) | ❌ refuses, *"context blocks [1]-[6] do not mention F004"* |
| `{"question":"fault F004"}` | ✅ *"For fault F004 … UnderVoltage …"* |
| `{"question":"F004"}` | ✅ answers |

So **the F004 chunk exists and is retrievable**, and `Allen-Bradley→Rockwell` **already exists** in `vendor-relevance.ts` — aliasing was never the cause. The real cause: the verbose query's extra tokens (`allen`, `bradley`, `powerflex`, `525`, `showing`, `mean`…) dilute the BM25 OR-fallback ranking so the F004 chunk drops below top-6 and the model **honestly** refuses.

## Fix (narrow, additive)

- **`extractFaultCodes()`** — detect single-letter + 2–4-digit codes (`F004`/`E001`/`A002`); deliberately skips two-letter model prefixes (`PF525`/`GS10`) and bare digits (`525`). Known limit: single-digit codes (`F4`) rely on the normal path.
- **`retrieveManualChunks()`** — when the query names a code, run one extra pass keyed on the code alone (the terse form that already retrieves the row) and merge those chunks **ahead** of the main BM25 result (dedup, cap topK). **Never removes a main-pass chunk, never causes a refusal.** Mirrors the engine's `recall_fault_code` *intent*, not its RRF/structured-table apparatus.
- **`isRefusalAnswer()` + route gate** — suppress phantom citation cards when the model refuses (chunks render 1:1, so a refusal otherwise ships with 6 citations — the other half of the contradiction).

Refusal logic is **not** weakened globally; the cross-vendor `stripConflictingVendors` strip is unchanged and still runs on the merged set (leak-guard test added).

## Evidence

- **+10 tests** (`extractFaultCodes`, code-pass prioritization, dedup, no-op on code-free queries, `isRefusalAnswer`, cross-vendor F004 leak guard).
- `npx vitest run src/lib` → **417/417 green**; `eslint` clean on all 4 files; **no new `tsc` errors** (the 7 pre-existing are in untouched `namespace/tree` + `rls-deny` test files).
- Cannot verify live this session (PR→smoke→deploy-vps); the regression test is the durable proof. The live probes above document the **current broken** prod state.

Closes the Surface 2 false-refusal item in PR #1875. Diagnosis correction + provider-fallback + mig-048 notes pushed to the #1875 checklist branch.

### DIFF
--- a/mira-hub/src/app/api/quickstart/ask/route.ts
+++ b/mira-hub/src/app/api/quickstart/ask/route.ts
@@ -6,6 +6,7 @@ import { cascadeComplete, type CascadeMessage } from "@/lib/llm/cascade";
 import {
   retrieveManualChunks,
   buildGroundedContext,
+  isRefusalAnswer,
   type ManualChunk,
   type ManualSource,
 } from "@/lib/manual-rag";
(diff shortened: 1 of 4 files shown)


Mikecranesync:
Known limit (ship-safe, noted to pre-empt a future question): `isRefusalAnswer` is a substring match on the refusal sentinel, so a *hedged* answer that quotes the refusal sentence mid-reply would suppress **all** citation cards even if the model also cited real chunks. Conservative-fails-safe and low-frequency — fine to ship; not building for it.

Issue #1318, opened 2026-05-15
P0: factorylm.com + app.factorylm.com TLS handshake hanging — VPS web layer stuck


Mikecranesync:
## Symptom

Both `https://factorylm.com/` and `https://app.factorylm.com/` time out for all HTTP requests despite the underlying TCP socket being open.

Discovered 2026-05-15 evening via CI E2E smoke failures on PRs #1297 and #1314 (and likely all subsequent PRs until resolved).

## Independent verification from CHARLIE (192.168.1.12, 2026-05-15 ~15:00 ET)

| Check | Result |
|---|---|
| `dig +short factorylm.com` (system + Cloudflare `@1.1.1.1`) | both `165.245.138.91` (DNS healthy) |
| Tailscale `ping` to Alpha | pong via LAN ~6 ms (CHARLIE network is fine) |
| `nc -z 165.245.138.91 443/80/22` | all OPEN — TCP layer up |
| `curl -v https://factorylm.com/` | TLS handshake stalls **after Client Hello**; 12 s timeout |
| `curl -v http://factorylm.com/` | request sent, **0 bytes received**, 8 s timeout |
| Same checks against `app.factorylm.com` | identical TLS-handshake stall |

## Diagnosis (likely)

The VPS at `165.245.138.91` accepts TCP but the web layer (nginx and/or its upstreams) is not responding to requests. Independent of DNS / GitHub Actions / CHARLIE network.

Probable root causes (operator to investigate):
- nginx upstream timeout — app containers behind it not responding
- Disk full on the DO droplet (eval-fixer / photo pipeline / log buildup)
- OOM killed nginx or one of its upstream services
- A container crashed — `mira-mcp` was already known missing (#1304); maybe more now
- iptables / UFW reload blocked traffic

## Investigation commands

```bash
ssh <prod-user>@165.245.138.91
docker ps -a
docker logs --tail=200 nginx 2>&1 | head -50
docker logs --tail=200 mira-pipeline 2>&1 | head -50
df -h
free -m
systemctl status nginx
docker compose -f docker-compose.saas.yml ps
journalctl -xe --since "1 hour ago" | grep -E "error|fail|oom|killed" | head -40
```

## Demo-day impact

**Blocks May 21 Florida Automation Expo demo** if not resolved — the iPad demo page (#1317, PR being merged) hits `app.factorylm.com/api/demo/*` endpoints. Demo target is T-6 days from filing.

## What CI sees

[...]

PR #1933, opened 2026-06-13
fix(upload): accept large PDFs end-to-end (raise size cap 20->50 MB)

Mikecranesync:
## Problem

Large PDFs (e.g. the 31.5 MB GS10 manual, the 33 MB Rockwell ref manual from ADR-0019) cannot be uploaded to the Hub. A Telegram-era 20 MB cap (Telegram limits PDFs to 20 MB) was copied onto **both** the Hub and the `mira-ingest` service, plus a 10 MB cap on the namespace node-files route.

## Root cause

The cap, duplicated across layers:

| Layer | File | Old cap |
|---|---|---|
| Hub local/folder upload | `local-upload.ts` | 20 MB |
| Hub `streamToBlob` (forward) | `mira-ingest-client.ts` | 20 MB |
| Hub cloud-pick | `api/uploads/route.ts` | 20 MB |
| Hub node-file attach | `namespace/node/[id]/files/route.ts` | 10 MB |
| Ingest service | `mira-ingest/main.py` `/ingest/document-kb` | 20 MB |

Critically, the **Hub and ingest caps had to move together**: the Hub accepting a file that ingest then rejects (its own 20 MB cap) would mark the upload `failed` downstream. Fixing only the Hub would look fixed but still fail.

## Change

- **Single source of truth** (`lib/config.ts`): `MAX_UPLOAD_MB` / `MAX_UPLOAD_BYTES` (`NEXT_PUBLIC_MAX_UPLOAD_MB`, default **50**). Replaces every hardcoded 20/10 MB cap, plus the client pre-checks and copy (`CreateChildCard`, `UploadPicker`, documents + manuals pages).
- **Ingest** (`mira-ingest/main.py`): `/ingest/document-kb` cap is now `MIRA_MAX_UPLOAD_MB` (default 50). **Must stay in sync** with the Hub's `NEXT_PUBLIC_MAX_UPLOAD_MB`.
- **Un-swallow the `formData()` parse error**: log the real reason + content-length before returning `invalid_multipart`, so a truncated/malformed-body failure is diagnosable instead of guessed at.
- **Rename** the now-wrong error key `exceeds_20mb_limit` → `exceeds_size_limit` (+ `limitBytes`); both client consumers updated.
- **Tests**: `mira-ingest/tests/test_ingest.py` locks the cap (accepts 31.5 MB, rejects > default 50 MB, honors the env override).

## Why 50 MB, not 100 / unlimited

`mira-ingest` buffers the whole file in memory (`await file.read()`) before docling — the documented 8 GB-VPS OOM path (`project_vps_oom_docling_incidents` / ADR-0019). 50 MB clears real OEM manuals at a bounded per-file footprint. **Host-OOM-safe**: since PRs #1318/#1319/#1336 every container is `mem_limit`-bounded (docling 3g, ingest 1g, OW 1.5g), so a large upload can at worst fail a single docling job inside its cap — not take down the host as in the May outages. True any-size needs the streaming ingest-v2 rebuild (ADR-0019), which is unbuilt. To go higher, raise both env vars (and nginx `client_max_body_size`, currently 100M).

## Investigated and ruled OUT

Next 16's 10 MB request-body clone ceiling (`experimental.proxyClientMaxBodySize` / `DEFAULT_BODY_CLONE_SIZE_LIMIT`) does **not** truncate route-handler uploads. Verified against the real standalone `server.js` (plus `next start`, `next dev`, middleware-matched + unmatched routes): a 31.5 MB multipart POST arrives intact even at the default 10 MB ceiling. No `proxyClientMaxBodySize` change is included.

A prior session reported the failure as `invalid_multipart` rather than the size cap; that could not be reproduced locally as Next truncation (leading hypothesis: an nginx/client timeout on a slow large upload). **After deploy, retry the 31.5 MB upload once** — the new `[local-upload] formData parse failed` log line will name the real cause if anything beyond the cap is involved.

## Testing

- `tsc --noEmit` clean on all edited files.
- `pytest mira-ingest/tests/test_ingest.py` → 10 passed (7 existing + 3 new cap tests).
- 5-config runtime verification of the ruled-out Next truncation (standalone `server.js` included).

## Out of scope (pre-existing)

- Unbounded ingest **concurrency** to docling (this bounds per-upload size, not N concurrent uploads).
- `namespace/node/files` stores bytes as Postgres `bytea`, so its raise means a single insert can write a 50 MB blob to Neon (insert-cost profile, not streaming).
- Streaming ingest-v2 for true any-size (ADR-0019).

### DIFF
--- a/mira-core/mira-ingest/main.py
+++ b/mira-core/mira-ingest/main.py
@@ -925,12 +925,18 @@ async def ingest_document_kb(
             detail=f"Only PDF files are supported (got mime={mime or 'unknown'}, ext={ext or 'none'})",
         )
 
-    # Validate size (20MB Telegram limit, enforced universally)
+    # Validate size. The old 20 MB cap was a Telegram-era artifact; real OEM
+    # manuals (31.5 MB GS10, 33 MB Rockwell ref) exceed it. Bounded — NOT
+    # unbounded — because this handler buffers the whole file (`await file.read()`
+    # above) before forwarding to docling, the documented 8 GB-VPS OOM path
+    # (ADR-0019). MUST match the Hub's NEXT_PUBLIC_MAX_UPLOAD_MB to avoid the
+    # Hub accepting a file ingest then rejects. True any-size needs ingest-v2.
     MB = 1024 * 1024
-    if len(raw) > 20 * MB:
+    max_upload_mb = int(os.getenv("MIRA_MAX_UPLOAD_MB", "50"))
+    if len(raw) > max_upload_mb * MB:
         raise HTTPException(
             status_code=413,
-            detail=f"File exceeds 20MB limit ({len(raw) // MB}MB)",
+            detail=f"File exceeds {max_upload_mb}MB limit ({len(raw) // MB}MB)",
         )
 
     form_tenant = (tenant_id or "").strip()
--- a/mira-core/mira-ingest/tests/test_ingest.py
+++ b/mira-core/mira-ingest/tests/test_ingest.py
@@ -175,3 +175,41 @@ def test_health_endpoint_returns_200(client):
     response = client.get("/health")
     assert response.status_code == 200
     assert response.json() == {"status": "ok"}
+
+
+# ---------------------------------------------------------------------------
+# document-kb upload size cap (MIRA_MAX_UPLOAD_MB, default 50)
+# Locks the 2026-06-13 fix: the Telegram-era 20 MB cap is raised to 50 MB and
+# made env-configurable, and MUST stay in sync with the Hub's
+# NEXT_PUBLIC_MAX_UPLOAD_MB. The size gate rejects before any Open WebUI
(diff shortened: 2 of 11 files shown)


Mikecranesync:
Superseded by #1935 (streaming ingest-v2 Slice 1), which cherry-picks this cap work (kept at 50 MB) and adds the memory-bounding (batched multi-row inserts + a NODE_INGEST_CONCURRENCY guard) plus an honest rewrite of the cap rationale. Leaving this open for you to close/merge as you prefer.

Mikecranesync:
Closing as verified-stale: Upload cap fix is on main: mira-hub config.ts MAX_UPLOAD_MB defaults to 50 (verified), synchronized with ingest.

Verified against current origin/main before closing (mechanism-level check, not title match). Branch preserved — reopen if any slice is still wanted. (Authorized backlog-hygiene sweep, 2026-07-28.)

Issue #1903, opened 2026-06-12
P1(security): node-attachment chunks inserted without is_private — potential cross-tenant exposure (verify after #1899)


Mikecranesync:
## Summary (needs verification — blocked by #1899)
The node=brain chunk writer inserts per-tenant manual chunks into `knowledge_entries` **without setting `is_private`**. The column defaults to `false`. Per `.claude/rules/knowledge-entries-tenant-scoping.md`, a per-tenant upload left `is_private=false` is exposed to **every** tenant through the hybrid read filter `(is_private = false OR tenant_id = $caller)` and the universal OEM/library aggregate surfaces.

## Where
`mira-hub/src/lib/node-knowledge-ingest.ts` → `writePdfChunksForNode()` INSERT INTO `knowledge_entries (... )` — no `is_private` column in the insert; relies on the DB default (`false`).

Contrast: `mira-hub/src/app/api/documents/upload/route.ts` correctly sets `is_private = true` with an explicit comment.

## Risk
A customer's private uploaded manual chunks could surface in other tenants' library/aggregate views (the #1833 leak class). Node-chat retrieval is subtree-scoped so it may not leak *there*, but the universal `is_private=false` surfaces would.

## Why "needs verification"
In my live test the upload 500'd (#1899), so **no row was actually inserted** — I could not confirm the leak end-to-end. This is a code-read finding. Confirm against staging once #1899 is fixed: upload a manual to a folder as tenant A, then check it does NOT appear for tenant B in `/api/knowledge`, `/api/library/*`, `/api/documents`.

## Fix
Set `is_private = true` in the `writePdfChunksForNode` insert (per-tenant node attachments are private by definition), and add a regression test asserting cross-tenant invisibility.

Mikecranesync:
✅ Shipped (mira-hub v2.2.5) + verified. `writePdfChunksForNode` now pins `is_private = true`; migration 052 backfilled existing v2 node_attachment chunks on staging + prod. Read-only check on staging: **0** leaky node chunks remain `is_private=false`, shared OEM corpus (83,532 public rows) untouched. Deployed to prod.

Issue #1899, opened 2026-06-12
P0(hub): folder=brain manual upload returns 500 — blocks beta gate (stranger can't attach own manual)


Mikecranesync:
## Summary
A brand-new ("stranger") customer cannot attach their own manual to a namespace folder. `POST /api/namespace/node/{id}/files/` returns **HTTP 500**, the file never attaches, and **no error is shown in the UI** — it silently looks like nothing happened. This blocks the beta gate end-to-end.

## Customer-journey repro (live prod, 2026-06-12)
1. Fresh signup at `https://app.factorylm.com/signup/` (email+password, instant, 8-day trial). New tenant "Dale Tech / owner".
2. Namespace → `Inbox` folder → **Upload one** → choose a PDF (2 KB synthetic service manual, `application/pdf`, well under the 10 MB cap).
3. Network: `POST https://app.factorylm.com/api/namespace/node/f7bbf542-3add-4983-94df-f82932f4c22d/files/` → **500**.
4. Folder still shows **"No files attached · 0 files"**. No toast, no error message.
5. Open the folder's **Ask MIRA** → ask *"My Zephyr ZX-9000 hydraulic press throws fault ZX-451…"* → MIRA replies *"I don't have manual coverage for the Zephyr ZX-9000 … [1]"* — correct (no hallucination), but only because the upload failed.

**Result: stranger uploads own manual → cited answer = FAILS.** This is exactly the North Star beta gate.

## Root cause (code path)
`mira-hub/src/app/api/namespace/node/[id]/files/route.ts` POST → `ingestPdfToNode()` in `mira-hub/src/lib/node-knowledge-ingest.ts`. The catch at route L173-176 returns `{"error":"Upload failed"} 500`, so `ingestPdfToNode` threw. Candidates (need prod logs to pin):
- `createUpload()` writing `hub_uploads` (table/column drift on prod).
- `unpdf` `getDocumentProxy`/`extractText` throwing on the PDF.
- `writePdfChunksForNode` INSERT into `knowledge_entries` — the v2 insert uses `source_type='node_attachment'`, `doc_id`, `ingest_route`, `page_start/page_end`, generated `content_tsv`; a missing column / constraint on prod throws.

Note: the **global** Knowledge upload path (`/api/documents/upload`, single-chunk) succeeded in the same session ("Parsed"), so this is specific to the node/folder v2 path.

## Fix asks
1. Pin the throw with prod logs; fix the failing dependency (schema or extraction).
[...]

PR #1935, opened 2026-06-13
feat(hub): memory-bounded node-attachment chunk writes (streaming ingest-v2 Slice 1)

Mikecranesync:
## Summary

Slice 1 of streaming ingest-v2: makes the as-built in-Hub PDF path
(`writePdfChunksForNode`, the single v2 chunk writer behind "folder = brain"
uploads, #1592/#1806/#1903) **memory-bounded**, and **supersedes #1933** by
cherry-picking its upload-cap single-source-of-truth.

### What changed
- **Bounded chunk accumulation.** Was: extract all pages eagerly → build one array
  of *every* chunk → one INSERT per chunk. Now: stream page-by-page, release each
  page's text once chunked, buffer chunks and flush each batch as **one multi-row
  INSERT** (`BATCH_ROWS=50`). Bounds both the rows held at once and DB round-trips.
  `tenant_id`/`source_url`/`doc_id` are fixed leading params; only id/content/page/
  metadata vary per row.
- **Concurrency guard.** `NODE_INGEST_CONCURRENCY` (default **1**) serializes parses
  via a race-free slot semaphore (a releaser with a waiter hands its slot over
  without decrementing). This contains the **dominant** memory term — the full file
  buffer + eagerly-extracted text — which the batching does *not* bound.
- **Cap = 50 MB (from #1933), honestly documented.** Kept at 50; rewrote the
  `config.ts` + `mira-ingest/main.py` comments to stop overclaiming that ingest-v2 is
  "memory-safe at any size." `50 MB` is a **policy** cap, not architectural.

### Honest memory accounting
The batching bounds the *smallest* term (chunk rows). The peak is still the file
buffer (≤2× via the pdfjs-safe `Uint8Array` copy) + the pdfjs proxy + eagerly
extracted `pages`. At default concurrency 1 the peak is **one** in-flight PDF, which
is what keeps the 8 GB VPS (ADR-0019) safe — not the batching. Raising the cap or
`NODE_INGEST_CONCURRENCY` raises the ceiling (`concurrency × O(file size)`); don't.

### Deferred → Slice 2
True per-page streaming extraction (`pdf.getPage(i).getTextContent()` instead of
eager `extractText`, drop the full-file copy) makes the text term `O(one page)` and
*earns* a cap toward nginx's 100 M. Plan: `docs/plans/2026-06-13-streaming-ingest-v2-slice1.md`.

## Test plan
- `npx vitest run` 3 specs under `mira-hub/src/lib/__tests__/` — **3 passed**:
  - `…-is-private.test.ts` — #1903 `is_private = true` invariant holds against the new
    batched SQL (assertion rewritten shape-independently of param numbering).
  - `…-batching.test.ts` — 60 single-chunk pages → exactly **2** inserts (50 + 10 tuples),
    bounded param counts, return = generated chunk count.
  - `…-concurrency.test.ts` — default concurrency 1 serializes the second parse.
- `ruff check mira-core/mira-ingest/main.py` — All checks passed.

Note: `unpdf@^1.6.2` is a declared dep (package.json + both lockfiles); a stale local
`node_modules` (pre-#1592) makes a local `tsc` report a spurious "cannot find module
'unpdf'" — CI/Docker install from `package.json`.

Supersedes #1933.


Mikecranesync:
## Verification notes (two gaps to know, neither blocks — documenting for the reviewer)

**1. The new multi-row SQL has no real-DB coverage in CI.** All three new specs mock
`c.query`, so they assert insert *shape* (tuple/param counts, `is_private` literal,
serialization), not real-DB execution. The node-attach route test mocks
`ingestPdfToNode` entirely, and Write-Path Integration (`tests/integration/test_write_paths.py`)
is Python HTTP-endpoint coverage — neither executes this TS INSERT against Postgres.
So its first real run is production. Mitigating factors: the SQL is verified correct by
inspection (placeholder math `$4,$1,'node_attachment',$5,$2,$6,$3,'v2',$6,$6,$7,true`;
`source_page=page_start=page_end=page`; `chunk_index` distinct per row so intra-batch
`ON CONFLICT DO NOTHING` can't self-collide; 203 params ≪ 65535), and on any SQL defect
the path **degrades to the legacy OW fallback** (the `try/catch` in `local-upload.ts`)
rather than hard-failing — i.e. it fails *quiet*, not loud. A real-pg vitest integration
test is a reasonable follow-up if we want loud failure here.

**2. The parse is awaited synchronously on the HTTP request path.**
`src/app/api/namespace/node/[id]/files/route.ts:178` does `await ingestPdfToNode(...)`,
so with `NODE_INGEST_CONCURRENCY=1` a second large-manual upload **queues** behind an
in-flight parse and could hit a gateway timeout under concurrent large uploads.
This is the deliberate latency-vs-OOM tradeoff: serializing parses is the right safety
default on the 8 GB VPS (one in-flight PDF, not N). If concurrent large uploads become
real, the fix is to background the parse (the `status:"parsing"` → polling pattern
already supports it), not to raise concurrency. No code change here — flagging so a 504
under that load isn't a surprise.

Full Hub vitest suite: **506 tests pass**; the one failing *suite* (`analysis.test.ts`,
`graphology` not found) is a stale-local-`node_modules` artifact (same class as the
`unpdf` resolve noise), pre-existing and unrelated — CI installs it from the lockfile.

Mikecranesync:
## Real-DB gap from the verification notes above — now CLOSED ✅

Verified end-to-end against the **dev** Neon branch (never prod) by running the *real* `ingestPdfToNode` + `retrieveNodeChunks`, not mocks. Added a reusable, self-cleaning harness (`mira-hub/scripts/bench-ingest-v2.ts`) + runbook (`docs/runbooks/ingest-v2-verify-and-benchmark.md`).

**6/6 PDFs (11 KB–967 KB, 3–27 pages) PASS:** rows-in-DB == returned chunkCount, every row `ingest_route='v2'` + `is_private=true`, every `chunk_index` distinct (the 73-chunk file crosses the BATCH_ROWS=50 boundary → two flushes, all 73 land). Steady-state ≈ 60–80 chunks/s, ≈ 750–1130 KB/s (first call is a cold-start outlier). All test rows cleaned up (0 left).

**Citation works:** `retrieveNodeChunks` ranks the correct source per query (e.g. 'how do I build a UDFB' → the UDFB_Intro doc p7; 'motor starter rung' → Conv_Simple_Complete p15), with filename + page.

**Still open (documented in the runbook §5):** the HTTP/next-auth upload layer and the final LLM-generated *cited answer* aren't covered by this harness; the synchronous-parse latency tradeoff stands.

Mikecranesync:
## Full E2E gap from the runbook §5 — now CLOSED ✅ (5/5 monitored runs)

The two items the DB-level benchmark left open (HTTP/auth layer + LLM-written cited answer) are now proven by `mira-hub/tests/e2e/ingest-v2-e2e-5run.spec.ts`, run on the **standalone build** (`node .next/standalone/server.js` — the #1899-fidelity shape `next dev` can't reproduce; Next 16's `next start` is broken under `output: standalone`).

Each run: **real credentials login → real upload SCREEN** (`namespace-file-input`.setInputFiles, not the API) → poll DB for v2 chunks → **Ask MIRA in the UI** → assert a citation chip from the **live cascade** + capture prose. Verdicts split infra-flake from path-failure (Gemini blocked → really Groq→Cerebras).

**Result (dev, 2026-06-13): 5 PASS / 0 PROVIDER_FLAKE / 0 PATH_FAIL in 37.5s.** Every run uploaded through the screen (1 chunk) and got a grounded answer with a `[1]` marker + chip:

> "The cause of Fault ZX-451 is the PT-7 pressure transducer drifting outside the 4-20 mA calibration band … **[1]**. To fix it: 1. Recalibrate the PT…"

All test tenants/rows cleaned up (0 left). Recorded in runbook §7. **Still open:** concurrency stress (runs were serial) and a large-file run through the UI (fixture is the 2 KB beta-gate PDF; the multi-batch path is proven at DB level in §4).

PR #2076, opened 2026-06-17
security(hub): fix /api/knowledge/search cross-tenant content leak

### LINKED ISSUE #1833: fix: Tenant-scope drift on /api/documents route

Mikecranesync:
## Summary

`GET /api/knowledge/search` added in #2044 had no `is_private` predicate. Per-tenant manual uploads now land with `is_private = true` (migration 052, #1903), so any authenticated tenant could read another tenant's private uploaded-manual snippet + source_url via this endpoint.

**Fix:** add `AND is_private = false` to both the BM25 query and the ILIKE fallback.
- Keeps universal shared-OEM-corpus search intact
- Does NOT add `tenant_id = $caller` (avoids #1761 "OEM returns 0 rows" regression)
- Matches `.claude/rules/knowledge-entries-tenant-scoping.md` and `/api/documents` reference impl

Patch source: `wiki/orchestrator/patches/2026-06-16-B9-knowledge-search-is-private.patch`

## Verification

```bash
# Static: 2 hits expected
grep -n "is_private = false" mira-hub/src/app/api/knowledge/search/route.ts

# Live test (if Neon staging available):
# 1. seed one is_private=true row for tenant A
# 2. search as tenant B for a term in A's content → expect 0 A-rows returned
```

## Test plan

- [ ] `grep -n "is_private = false" mira-hub/src/app/api/knowledge/search/route.ts` — expect 2 hits
- [ ] Staging-gate passes (no BM25/search regressions on OEM corpus)
- [ ] Knowledge → Manuals search still returns OEM results for known terms (e.g. "GS10", "PowerFlex")

Closes #1833 (the cross-tenant leak class on this route)


Mikecranesync:
Superseded by #2127 which landed the same `is_private = false` guard on both the BM25 and ILIKE queries. Fix is already on main (commit 529d62e2).

Issue #2212, opened 2026-06-21
P2 corpus(rag): seed common non-GS10 OEM VFD manuals (Danfoss/ABB/Mitsubishi/SEW/Lenze/SINAMICS/Pilz)


Mikecranesync:
**Found by** retrieval-grounding investigation (2026-06-21). The corpus-coverage (not code) half of the ungrounded set. Broader than #2198 (which is the GS20 instance).

## Finding
Bucket-(c) ungrounded questions cluster on **non-GS10 OEMs** the KB doesn't carry — MIRA correctly falls back to general knowledge because the manual genuinely isn't seeded. Real failing questions from the export:
- Danfoss VLT (alarm 14 earth fault, alarm 10 ETR, FC102 alarm 4 mains phase loss, AQUA FC202)
- ABB ACS355/ACS150 (motor overload, single-phase input)
- Mitsubishi FR-D720 / A700 (OC on startup, Pr.7)
- SEW MOVITRAC, Lenze i550, Siemens SINAMICS G120, Pilz PNOZ (×22)

## Action
Seed these OEM manuals to the **shared** KB (staging first, verify BM25 recall, then prod via `seed-oem-manuals.yml` — never prod-first). Prioritize by frequency in the export.

## Note
The corpus is GS10/garage-conveyor-centric today. This is the "stranger's equipment" gap that directly limits marketplace breadth. Pairs with #2198 (GS20) and the best-effort-same-manufacturer UX question there.

Issue #2211, opened 2026-06-21
P2 bug(rag): retrieval extraction precision — vendor false-positive full-suppress, hardcoded product allowlist, weak model exclude


Mikecranesync:
**Found by** retrieval-grounding investigation (2026-06-21). Three related precision defects in `mira-bots/shared/` extraction (all pure-function, offline-testable). Related to #2083 (cross-vendor false-DROP, resolved) and #2095 (product quota) but distinct mechanisms.

## (1) Vendor false-positive → full suppression [P2]
`rag_worker.py:623-654` cross-vendor filter uses a vendor from `uns_resolver`. A false-positive alias hit ("delta pressure too high" → `VENDOR_ALIASES["delta"]` → Delta Electronics, `uns_resolver.py:110`) makes `chunk_matches_vendor` (`:62-88`) drop every correct chunk; when nothing survives it **fully suppresses** (`:646-654`) → 0 chunks.
- Fix: apply the cross-vendor filter only at vendor confidence ≥0.7, or require whole-word (not English-substring) vendor match; tighten high-collision aliases (`delta`, `ab`, `sew`).

## (2) Product-rerank allowlist is hardcoded → never fires for most gear [P2]
`_PRODUCT_NAME_RE` (`neon_recall.py:78-97`) omits families already in the UNS alias table (MOVITRAC, MOVIDRIVE, FR-E/A/D, V1000/J1000/GA500…) and **can never match a stranger's uploaded equipment**. So the product-name rerank stream (surfaces the most relevant chunk from the right manual) silently never runs.
- Fix: derive product-name candidates from `state["uns_context"].model/product_family` (already extracted), not a second hardcoded regex.

## (3) `_product_search` exclude only blocks the "0" suffix [P3]
`neon_recall.py:392-393` `exclude_pat = f"%{name}0%"` blocks "PowerFlex 40"→"400" but not 401–409/40A.
- Fix: word-boundary model match (`model_number ~ '(^|[^0-9])40([^0-9]|$)'`).

All three: pure functions, unit-testable with stub chunks, no DB.

Mikecranesync:
**Fully complete.** Sub-fixes 1 & 2 (vendor-confidence gate + product-hint fallback) shipped in PR #2913 (v3.213.1); sub-fix 3 (model-suffix exclusion widening) shipped in **PR #2931** (`7931d98dc`, v3.216.3, deployed). All verified with regression tests on origin/main. Closing as completed.

Issue #2210, opened 2026-06-21
P0 bug(rag): bot recall path lacks the hybrid is_private filter + request-tenant — tenant uploads invisible to the bot


Mikecranesync:
**Found by** retrieval-grounding investigation (2026-06-21). Distinct from #2112 (that's the Hub `/api/knowledge/search` leak); this is the **bot recall path** (`mira-bots/shared/neon_recall.py`), which the tenant-scoping ⏳ rollout list does NOT cover.

## Bug
Every recall query filters `(tenant_id = :tid OR tenant_id = :shared_tid)`:
- `neon_recall.py:712, 361, 408, 508, 880, 959`; `:tid` comes from `rag_worker.tenant_id` / `ask_api` = the **deploy-wide `MIRA_TENANT_ID` env** (`rag_worker.py:394`).

In prod `MIRA_TENANT_ID == SHARED_TENANT_ID` (`docs/AUDIT.md:199`, `neon_recall.py:104`) — both bind params are the same UUID, so the path only ever sees the **shared OEM/system corpus**. A stranger's uploaded manual is stored under *their own UUID* with `is_private=true` (per `.claude/rules/knowledge-entries-tenant-scoping.md`) → **invisible to bot recall**.

## Scope / honesty
- In prod the bot is effectively single-tenant, so this bites only on **multi-tenant bot surfaces**. If beta strangers ask via **Hub chat** (which has its own hybrid filter) this may not block them; if they ask via Telegram/kiosk bound to their tenant, it does. **Confirm which surface beta uses before prioritizing the full fix.**

## Fix
- Proper: replace the predicate with the canonical `(is_private = false OR tenant_id = :caller)` and thread the **request** tenant (not `MIRA_TENANT_ID`) from the adapter.
- Short-term mitigation: `(is_private = false OR tenant_id = :tid)` so the shared corpus is selected by `is_private`, not by an env UUID that happens to equal the system tenant.

## Offline-testable: partially
Predicate-shape + param-binding unit test (assert WHERE clause + params with a stub conn). Full behavior needs a DB.

Mikecranesync:
Verified fixed on origin/main: `mira-bots/shared/neon_recall.py` now applies the hybrid corpus filter `(is_private = false OR tenant_id = :tid)` in both retrieval streams (lines ~405/457), with request-tenant threaded via `:tid`. Closing as already-shipped. (Orchestrator liveness sweep 2026-07-13.)

Issue #2209, opened 2026-06-21
P1 bug(rag): multi-turn follow-up turns drop equipment context → 0 chunks (~34% of ungrounded)


Mikecranesync:
**Found by** retrieval-grounding investigation (2026-06-21), data forensics on 3,725 exported production traces. ~34% of ungrounded turns.

## Bug
Bare follow-up turns ("Haven't meggered it yet", "No recent changes, been running fine for six months", "Voltage at the MCC bus fluctuates between 440 and 485", "ok, what do you think is causing it?") embed with **no equipment signal** — the prior-turn VFD model / fault code is **not carried into the RAG query**, and `sessionId` is `None`. JSONL confirms `input.query` is only the follow-up sentence.

Result: a follow-up that doesn't restate "GS10 / overcurrent" retrieves nothing → ungrounded answer mid-diagnosis. **Routing gap, not a corpus gap** — independent of seeding more manuals.

## Fix
Carry conversation/asset context forward into the retrieval query (or session-scoped retrieval): when the turn is a follow-up in an active diagnostic session, prepend the resolved equipment (`state["uns_context"].manufacturer/model` + last fault code) to the recall query.

## Offline-testable: partially
The recall-query construction for a follow-up turn is unit-testable (assert equipment tokens are present in the query when state carries them); end-to-end grounding needs the live eval.

## Evidence
Representative: "Haven't meggered it yet" (×16), "No recent changes…" (×14), "Voltage at the MCC bus…" (×6) — all 0 chunks, sessionId None.

Mikecranesync:
Already implemented + merged in **PR #2913** (squash `1cc1ae3a6`, v3.213.1, now live on prod ≥v3.216.x). The issue was left open by oversight. Verified on origin/main today: regression tests present and green — #2208 `tests/test_fault_code_extraction_gate.py` (25 passed, proximity+shape fault-context gate on all 3 code shapes), #2209 `mira-bots/tests/test_multiturn_equipment_context.py` + `test_multiturn_context_e2e.py` (18 passed, engine.py enriched-query precedence at :250). Behavioural confirmation is the shared staging replay `docs/eval/rag-grounding-staging-replay-2026-07-25.md` (not yet executed — tracked separately). Closing as completed.

PR #2213, opened 2026-06-21
fix(rag): drop embedding-not-null filter from kb_has_pair_coverage (parity with #1308)

Mikecranesync:
## Why
From the 2026-06-21 retrieval-grounding investigation (issues #2207–#2212). This is the **safe, correct-by-construction** finding — the rest are filed for the staging-gated work.

`kb_has_coverage` deliberately dropped `AND embedding IS NOT NULL` in **#1308** (a BM25-reachable row is still coverage; seeded-but-unembedded rows were invisible → Modbus questions routed to the hallucination fallback). Its sibling **`kb_has_pair_coverage` kept the filter** (`neon_recall.py:962`). So a real `(vendor, model)` pair whose chunks are seeded but not yet embedded (the #2083/#2085/#2117 NULL-embedding class) is judged "not covered" → the resolver **drops the model** (`UNS_PAIR_DROPPED`), degrading the UNS path, the product-name rerank stream, and the citation label — for a product the KB *does* cover lexically.

## What
- Removed `AND embedding IS NOT NULL` from `kb_has_pair_coverage`, restoring parity with `kb_has_coverage`.
- Extracted the query to `_KB_PAIR_COVERAGE_SQL` so it's unit-testable without NeonDB.
- `tests/test_kb_pair_coverage_sql.py` — asserts no embedding filter + still pair/tenant-scoped.

## Verification
`pytest tests/test_kb_pair_coverage_sql.py` → 3 passed; `ruff` clean; module imports. No DB needed (the fix is a contract change, the sibling already proved the behavior in #1308).

## Deploy gate
Engine/RAG path — **run the staging eval (garage_conveyor_field --live, #2202) before prod deploy.** Low risk (aligns to an already-proven sibling), but it changes coverage judgments.

## Related (filed this session, not in this PR)
#2207 (dead 0.70-floor relaxation), #2208 (chat query poisoning), #2209 (follow-up context loss), #2210 (bot recall hybrid filter, P0), #2211 (extraction precision bundle), #2212 (non-GS10 corpus seeding).

### DIFF
--- a/mira-bots/shared/neon_recall.py
+++ b/mira-bots/shared/neon_recall.py
@@ -912,6 +912,21 @@ def kb_has_coverage(vendor: str, model: str, tenant_id: str) -> tuple[bool, str]
 
 KB_PAIR_COVERAGE_MIN_CHUNKS = int(os.getenv("MIRA_KB_PAIR_COVERAGE_MIN_CHUNKS", "1"))
 
+# No ``embedding IS NOT NULL`` filter — kept in parity with ``kb_has_coverage``
+# (#1308). A row reachable only via BM25 (content_tsv) is still KB coverage, and
+# a freshly-seeded (vendor, model) pair whose embeddings haven't been backfilled
+# yet must NOT be judged "not covered": that makes the resolver drop the model
+# (UNS_PAIR_DROPPED), degrading the UNS path, the product-name rerank stream, and
+# the citation label for a product the KB does cover lexically. See the sibling's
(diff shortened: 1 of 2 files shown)


Mikecranesync:
Superseded by the fresh re-cut just opened (worklist #2990 item P1-4) — the fix was re-applied onto the function's current hybrid-corpus-filter shape, with an offline contract test locking the parity. This branch predates that refactor.

Issue #2263, opened 2026-06-22
design(kb): manufacturer fragmentation in Knowledge — 83K chunks, 24K uncategorized, alias/fuzzy-merge needed


Mikecranesync:
## Problem

The Knowledge page shows severe manufacturer fragmentation across 83K chunks:
- **Deshazo**: ~10 spelling variants (DeShazo, Deshazo, DESHAZO, etc.)
- **Allen-Bradley / Alien-Bradley**: OCR artifact
- **Rockwell Automation** (34K chunks) vs **Rockwell** (18 chunks) — same vendor, split key
- **Automationdirect** vs **Automation Direct** — casing variant
- **Magnetek**: encoding artifact variants
- **~24K chunks (~78% of docs) are "Uncategorized"** — no manufacturer assigned

## Impact
Retrieval precision suffers when a query for "Allen-Bradley" misses chunks tagged "Rockwell Automation" or vice versa. The existing `manufacturer-aliases.json` at `mira-hub/src/lib/manufacturer-aliases.json` partially addresses this but is incomplete.

## Proposed approach
1. **Alias/fuzzy-merge**: Extend `manufacturer-aliases.json` with the known variants. Apply at query time (already done in PR #2083) AND at display time in the Knowledge page rollup.
2. **OCR artifact fix**: Normalise "Alien-Bradley" → "Allen-Bradley" in the ingest pipeline (`mira-crawler/ingest/`).
3. **Classifier pass on Uncategorized**: Run a LLM or regex classifier over the 24K uncategorized chunks to assign manufacturer from title/content. Backfill as a one-shot migration.

## Design deliverable
Document the chosen approach as an ADR or plan before any code changes.

## Constraints
- Follow `.claude/rules/uns-compliance.md` — UNS path builders in `mira-crawler/ingest/uns.py`
- Read filter `(is_private=false OR tenant_id=$caller)` must remain intact
- No auto-promotion of proposed entities

Mikecranesync:
Claude-worktree audit (2026-08-30): the stale `fix/manufacturer-fragmentation` tree from closed PR #2275 should not be transplanted. Its core work was superseded by merged PRs #2937 and #2995, and current `main` passes the cross-surface alias consistency test (2/2).

Fresh results on the stale tree:

- crawler normalization: 18 passed;
- Hub normalization: 6 passed;
- cross-surface consistency: 1 passed / 1 failed.

[...]

PR #2275, opened 2026-06-22
fix(kb): collapse manufacturer fragments in Knowledge rollup (#2263)

Mikecranesync:
## Summary

- Extends `manufacturer-aliases.json` with three new entries: `"rockwell"` → `"Rockwell Automation"`, `"automationdirect"` / `"automation direct"` / `"automationdirect.com"` → `"AutomationDirect"`
- Applies `normalizeManufacturer()` in `/api/knowledge` **after** the SQL `GROUP BY` to re-aggregate rows that map to the same canonical name (e.g. 34K "Rockwell Automation" + 18 "Rockwell" → single "Rockwell Automation" row)
- Syncs `OCR_VARIANT_ALIASES` in `mira-crawler/ingest/manufacturer_normalize.py` to stay in lockstep with the alias JSON (existing cross-surface consistency test guards this)

Partial fix for #2263. Remaining work: OCR artifact fix in ingest pipeline, classifier pass on 24K Uncategorized chunks.

## Root cause

The Knowledge page was grouped purely by SQL `INITCAP(LOWER(TRIM(manufacturer)))` with no alias pass, so "Rockwell Automation" and "Rockwell" produced separate catalog rows. The alias map existed but was only applied at upload time and in quickstart, not in the library rollup.

## Test plan

- [ ] `/api/knowledge` response: verify "Rockwell Automation" row count now absorbs the former "Rockwell" row (18 extra chunks)
- [ ] "AutomationDirect" appears as a single catalog row (no "Automationdirect" duplicate)
- [ ] Knowledge page manufacturer list is shorter by collapsed entries


Mikecranesync:
Superseded by the fresh re-cut just opened (worklist #2990 item P1-6). The alias-table half of this PR already landed via the manufacturer_normalize lineage; the missing route-side re-aggregation was re-applied onto current main. 

PR #1657, opened 2026-06-02
feat: Walker DT-2026 gap closure — Phases 0–5 (Connect→Collect→Store→Visualize)

Mikecranesync:
## Walker DT-2026 Gap Closure — Phases 0–5

Closes the gaps from the DT-2026 alignment analysis, governed by `docs/plans/2026-06-01-mira-master-architecture-plan.md` (PRIMARY FOCUS). Builds the **Connect → Collect → Store → Visualize** spine end-to-end.

> **Stacked PR.** Base is `feat/hub-command-center` (not `main`) because Phase 4 depends on the Command Center code that lives only on that branch. Merge the CC branch first, then this.

### Phases

| Phase | What | Evidence |
|---|---|---|
| **0 — Repo truth audit** | `docs/plans/current-state-gap-closure-plan.md` — 8-stage audit, every component status-labeled w/ file/PR evidence. Corrects 2 master-plan errors (HealthScoreWidget IS built; `/plc` page MISSING). | CodeGraph + verified-absent |
| **1 — Schema** | Hub migrations `032`–`036` + ADR-0022. `decision_traces`/`tag_events`/`flaky_input_signals`/`approved_tags` net-new; **`current_tag_state` = extend `live_signal_cache`** (reuse, not duplicate). | Applied to ephemeral pg16; idempotent; append-only verified |
| **2 — Ingest API** | `POST /api/v1/tags/ingest` (mira-relay): HMAC, fail-closed allowlist, UNS resolve, append `tag_events` + atomic upsert cache. `simulated` derived per-batch; sim never clobbers real; null reads rejected; events+cache in one `persist_batch` txn. | 66 pytest + real-pg end-to-end |
| **3 — Ignition collector** | `api/tags/collector.py` (pure, reuses signing+allowlist) + rewrote the previously-**unsigned** `tag-stream.py` → HMAC + Phase-2 contract + retry/backoff, read-only. + install doc + saas-compose env wiring. | 24 collector tests; Ignition suite 74/74 |
| **4 — Command Center freshness** | "Live" now = **tag freshness** (live/stale/unknown/simulated) from `current_tag_state`; HTTP reachability kept as a separate status. Migration-lag-guarded. | 16/16 bun pure tests; `tsc` clean across mira-hub |
[...]

PR #2283, opened 2026-06-24
fix(migrations): 057 — add ingest columns to tag_events (additive drift repair)

Mikecranesync:
## Summary
Additive migration repairing a **schema drift** on `tag_events` discovered during the SimLab→UNS ingest staging validation.

`mira-relay/tag_ingest.py` (the relay ingest landing) INSERTs `tag_events` with the columns from `033_tag_events.sql` — `value, value_type, quality, source_system, source_connection_id, equipment_entity_id, metadata`. On environments where an **earlier diff-style `tag_events`** pre-existed, `033`'s `CREATE TABLE IF NOT EXISTS` **skipped** it, so those columns were never added → landing fails with `column "source_system" does not exist`. This is the `CREATE TABLE IF NOT EXISTS` hazard in `.claude/rules/mira-hub-migrations.md`.

**Confirmed on staging 2026-06-24:** all 7 columns missing on `tag_events` (`live_signal_cache` was already correct, 0 missing).

## What it does
Purely **additive**: `ALTER TABLE tag_events ADD COLUMN IF NOT EXISTS` ×7, with the **exact** 033 types/defaults/CHECKs. No drop / rename / rewrite / truncate / backfill. Idempotent — a no-op where `tag_events` already matches 033. `source_system TEXT NOT NULL` (no default) is safe because the only drifted env has an **empty** `tag_events` (verified 0 rows on staging).

## Status
- **Dry-run on staging:** ✅ rollback-verified (columns appear in-txn, gone after rollback, no errors).
- **Applied to staging:** ✅ all 7 columns now present (done directly to unblock the validation).
- **Prod:** likely the same drift (shared migration lineage) — applying 057 via `apply-migrations.yml` is safe regardless (`IF NOT EXISTS`). Do **not** merge until reviewed; promote dev→staging→prod via the normal gate.

Not for immediate merge — opened as the durable artifact for the prod path.

### DIFF
--- a/mira-hub/db/migrations/057_tag_events_ingest_columns.sql
+++ b/mira-hub/db/migrations/057_tag_events_ingest_columns.sql
@@ -1,0 +1,53 @@
+BEGIN;
+
+-- Migration 057: backfill the ingest columns onto tag_events (drift repair).
+--
+-- WHY THIS EXISTS
+--   `mira-relay/tag_ingest.py` (ingest_batch -> NeonTagStore.persist_batch)
+--   INSERTs tag_events with the columns defined by 033_tag_events.sql:
+--     value, value_type, quality, source_system, source_connection_id,
+--     equipment_entity_id, metadata.
+--   On environments where an EARLIER, diff-style `tag_events` (event_type /
+--   prev_value / new_value / threshold / window_start|end / severity ...) was
+--   created before the current 033, `CREATE TABLE IF NOT EXISTS` in 033 SKIPPED
+--   the table, so those ingest columns were never added. The SimLab->relay
+--   ingest landing then fails with `column "source_system" does not exist`.
+--   (Confirmed on STAGING 2026-06-24: tag_events was missing all 7 columns;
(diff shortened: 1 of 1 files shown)


Mikecranesync:
Closing — superseded by the staging-only R1 reconciliation (2026-06-24).

Root cause was a **staging-only** drift: `migration-verify.yml` auto-applied PR #1657's *early-draft* 033/035 to the persistent staging branch; `CREATE TABLE IF NOT EXISTS` + the filename-keyed ledger then froze staging at the draft shape. **Prod was already canonical** (verified read-only via db-inspect: tag_events=033, approved_tags=035 with composite PK + 58 real rows); dev had the tables absent.

So 057 (additive columns on tag_events) is a no-op on prod and insufficient for staging (which also had extra NOT-NULL draft columns + wrong PK). Staging was reconciled by dropping the two **empty, orphaned** tables and re-running canonical 033/035. The full 8-step ingest validation is now **8/8 GREEN** against staging. Plan: `docs/plans/2026-06-24-ingest-schema-reconciliation-plan.md`.

Issue #2691, opened 2026-07-14
drive-pack-extract: add a Magnetek/IMPULSE extractor dialect (0/0 on G+ Mini — mnemonic faults + dotted params + int-keyed schema)


Mikecranesync:
## Summary

The Drive Commander scout ran the real **Magnetek IMPULSE G+ Mini** manual (144-25085, firmware 14515) through the production pipeline in PR #2690 and recovered **0 fault codes / 0 parameters** — grade B 85.7 INCOMPLETE / NOT PROMOTABLE. The scout correctly headlined it as a generalization gap. This issue tracks the real fix: a Magnetek extractor dialect.

## Root cause (two layers)

1. **Layout:** the G+ Mini fault table is a 3-column `Fault | Name/Description | Corrective Action` grid with **mnemonic codes** (`dnE`, `EF0`–`EF8`, `LF`, `LL1`/`LL2`, `oC`, `UV1`, `BE2`/`BE3`/`BE6`, `MNT`). Parameters are **dotted** (`H01.01`, `U01.10`, `C12.05`). The position-aware parser in `extractor.py` is tuned to the PowerFlex 520/525 + PowerFlex 40 table shapes.
2. **Schema:** `mira-bots/shared/drive_packs/schema.py` `LiveDecode.fault_codes` is `dict[int, str]`. Magnetek/DuraPulse faults are **mnemonic**, so even a correct parse can't key them into `fault_codes` as-is. This is the same structural gap behind GS10/GS20 (#2685).

## Proposed work

1. Read the real fault-table (~pp.136-140) + parameter pages first — don't guess (`superpowers:systematic-debugging`).
2. Add a Magnetek/Yaskawa-V1000 dialect branch to `extractor.py` (3-column mnemonic fault table + dotted-param layout), mirroring how PF40 was added alongside 520/525.
3. Decide the mnemonic-fault representation: either extend the schema to allow string-keyed fault codes, or a documented mnemonic→code mapping. Coordinate with the GS10/GS20 mnemonic gap (#2685) — likely one shared fix.
4. Add a `gold/magnetek_impulse_g_plus_mini` fixture + regression once a dialect lands.
5. Then the crane-safety **answer judge** + crane Q&A coverage cases (the answer-only half of the original task, N/A in #2690).

## Refs
- PR #2690 (this run + `magnetek` grading convention + crane pack-content check + history-driven scout)
- Related: #2685 (GS20 same class), GS10 gold-independent-only
[...]

PR #2692, opened 2026-07-14
docs(eval): Run-A frozen evidence baseline for G+ Mini fault-code schema (Refs #2691)

Mikecranesync:
## What this is (and is NOT)

This PR adds the **immutable Run-A evidence baseline** for the drive-pack fault-code schema
investigation on the IMPULSE **G+ Mini** (crane/hoist VFD, mnemonic fault codes `oC`/`BE2`/`LL1`).

- ✅ It **is** a frozen, hashed, independently-verified measurement of *current production behavior*.
- ❌ It is **NOT** the extractor fix / dialect. It makes **no** runtime schema decision.

**Refs #2691** (does **not** close it). #2691 — *"add a Magnetek/IMPULSE extractor dialect (0/0 on
G+ Mini — mnemonic faults + dotted params + int-keyed schema)"* — is the eventual fix; this PR is
the comparison floor that fix will be measured against.

## Commits

| Commit | Kind | Effect on evidence |
|---|---|---|
| `3afcb50d` | `docs(eval)` frozen baseline | the immutable Run-A artifacts |
| `a54c58fc` | `style(eval)` ruff F541 lint-only | **none** — see "lint does not affect evidence" below |

## Headline — the honest floor (not massaged)

| Metric | Value |
|---|---|
| Deterministic resolution | **0%** |
| Fallback | **0%** |
| Unresolved | **100%** |
| Hard failures (unsafe guesses) | **0** |

Under `fault_codes: dict[int, str]` (`schema.py:56`) + the `_int_keyed` gate (`loader.py:275`),
a source-preserving mnemonic-keyed G+ Mini pack **fails to load entirely** — not near-empty. The
empty pack is the **correct, honest result**, from two independent causes kept separate in the
report: (1) **no G+ Mini source material exists** anywhere (Magnetek isn't even a recognized
vendor), and (2) the int-keyed schema cannot represent a mnemonic identifier. Exercised against
the **real production loader**, not a reconstruction. GS10 abstraction leak documented (mnemonic
smuggled into the int-keyed *value* string, e.g. `'21' → 'oL overload'`).

## The lint cleanup does not affect the frozen evidence

`a54c58fc` only removes two extraneous `f` prefixes from placeholder-less strings in the manifest
generator. Therefore:

- manifest **output is byte-identical** (regenerating produced zero diff),
- `make_manifest.py` is **self-excluded** from hashing,
- every artifact **SHA-256 in `MANIFEST.json` is unchanged**,
- all **three read-only Sonnet verifiers remain PASS** (hash recompute, factual claims, safety invariants).

## Run B — the five locked constraints (design only; `docs/eval/drive-pack-faultcode-runA/RUN_B_DESIGN.md`)

1. **C1** — preserve the exact raw fault token verbatim (no casefold at capture).
2. **C2** — fallback may resolve identity/description/severity **only** with a manual citation.
3. **C3** — fallback may **not** invent codes, claim unverified wire/register values, or give
   reset/override/clearing/operational guidance.
4. **C4** — safety-relevant unresolved crane/hoist faults stay **unresolved and hard-fail** (never
   guessed); Run B must be **provably no-less-safe than Run A's 0 hard failures**.
5. **C5** — an approved fallback record must convert to a deterministic parser/pack entry + a
   `fixtures/` case + a regression test proving Claude is no longer needed.

## Scope guarantees (explicit)

- **No** runtime schema decision · **No** `gold/` promotion · **No** deployment · **No** flags ·
  **No** WO-evidence changes · **No** production effects.
- Raw input is a **clearly-labeled synthetic probe** — no manual content fabricated.

## For Run B

Run B must use **this PR's frozen artifacts as its comparison floor** — do **not** regenerate
Run A. The floor lives at `docs/eval/drive-pack-faultcode-runA/` (baseline commit `3afcb50d`).


Mikecranesync:
Superseded by a clean re-parent. The branch `chore/drive-pack-faultcode-runA-baseline` was mistakenly cut from an unmerged feature commit (`db1177c9` on `fix/ship-container-ops-agents`), so this PR's diff dragged in 4 foreign ops-agent files and a VERSION downgrade (3.140.2→3.127.1). Reopening from `docs/drive-pack-faultcode-runA-baseline`, parented on main with a docs/eval-only diff. The originally-frozen commits `3afcb50d`/`a54c58fc` remain intact on this branch as the immutable record; the new PR carries byte-identical artifacts (all SHA-256 unchanged).

PR #2819, opened 2026-07-19
fix(inference): reasoning-model handling — think-strip + bounded burn-retry in the cascade router (v3.176.5)

Mikecranesync:
## fix(inference): reasoning-model handling in the cascade router (v3.176.5)

**The measured failure this fixes (Tower OP round 4, PR #2818):** with `TOGETHERAI_VISION_MODEL=moonshotai/Kimi-K2.6`, 20/22 production calls returned HTTP 200 with the entire 1024/2000-token output budget burned on chain-of-thought and **zero visible content** → `EMPTY_RESPONSE` → deterministic fallback on all 12 bench cases — while the same model answered the hardest bench question photo-exactly on a raw call. The one-env-var vision flip is dead on arrival without this.

### What it does
- **`_strip_reasoning()`** on every provider reply: closed `<think>…</think>` blocks removed; an unterminated `<think>` (the cap-hit-mid-reasoning signature) yields no visible content. Also applied on the rate-limit inline-retry path.
- **Reasoning-burn recovery:** empty visible content + burn evidence (completion tokens at the cap, `reasoning_content` field, or think markup) → ONE retry against the same provider with `LLM_REASONING_RETRY_MAX_TOKENS` headroom (default **8192**, `0` disables, or-form parsed). Genuinely empty replies are NOT retried; a still-empty retry falls through to the existing cascade behavior.
- Compose mapping at every `TOGETHERAI_TIMEOUT` site (bot / staging / staging-vps / saas ×10 lines) + `docs/env-vars.md` row — the enumerated-compose-env-block trap honored.

### Evidence
- 18 new hermetic tests (`mira-bots/tests/test_router_reasoning.py`): strip semantics, burn detection (all three signals), single-retry with headroom, retry-disabled knob, no-retry-on-plain-empty, LFM-style strip-on-success.
- Existing suites green: `mira-bots/tests/test_router_coverage.py` 22 passed, root `tests/test_inference_router.py` 29 passed (run separately per the two-tree rule). ruff check+format clean; pyright 0 errors on changed files.
- Round-4 forensic detail in `tools/internet_print_test/benchmarks/2026-07-18-towerop/ROUND4-2026-07-19-together-qualification.md` (PR #2818).

R5 bench rounds on CHARLIE will run against this branch before any staging flip.

https://claude.ai/code/session_01VsSvoaumZz7y6E9yWJC1P3


Mikecranesync:
**Bench evidence (Tower OP rounds 5-beta/5-gamma, full report in PR #2818's branch — `ROUND5-2026-07-19-push-to-9-loop.md`):**

- Round 4 (same Kimi, no fix): **0/12 replies**, 2 classification misroutes. Beta (MiniMax + this fix): **10/12 real cascade replies, 12/12 correct classification**. Gamma logs: 21 live `REASONING_BURN` retries firing as designed.
- Raw-lane validation of the retry cap: Kimi's 5 alpha burn-outs all recovered at 8192 (outputs 3.6k–5.4k), composing to an **8.54/10** judged mean — the best config measured.
- Caveat for reviewers: through the *production theory template* the same models still burn past 8192 on 4 hard cases (prompt weight lengthens reasoning) — that's a prompt-side follow-up (slim answer-first variant), not a router issue. Production means: beta 5.25, gamma 4.46 vs 1.9 pre-loop baseline.

PR #2849, opened 2026-07-20
[TEST-DO-NOT-MERGE] CI Gate failure-mode verification

Mikecranesync:
Throwaway verification PR. Push 1 (evidence #5): the `changes` job is intentionally broken (`exit 1`) — CI Gate MUST go red (fail-closed: a failed changes job can never pass the gate). Push 2 will instead break a gated job (evidence #4). Injections removed before any real merge; branch deleted after.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -58,8 +58,6 @@ jobs:
               - '!**/*.png'
               - '!**/*.jpg'
               - '!**/*.gif'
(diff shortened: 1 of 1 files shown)


Mikecranesync:
Push 2 (evidence #4): reverted the changes-job break; now the gated **migration-order-check** job is intentionally broken (`exit 1`) while `changes` succeeds (code=true). CI Gate MUST go red because a gated always-run job did not equal success.

PR #2879, opened 2026-07-23
fix(dataset): harden paid gate message and evidence checks

Mikecranesync:
## Summary

Post-merge hardening follow-up to #2878 / PR 3. This closes the remaining adversarial-review holes in the dataset v0 paid gate before PR 4 starts Together orchestration.

- validates chat training messages before a record can become dataset-eligible
- rejects invalid messages during assembly with `MESSAGE_INVALID`
- re-checks hand-built datasets with `all_records_message_valid`
- makes invalid held-out lineage evidence block instead of raising
- requires non-future model-support timestamps plus a receipt reference
- derives `DatasetV0.source_systems` from eligible records instead of constructor state
- removes caller-overridable cost-estimator knobs

Pure code only: no network calls, no spend, no upload, no training, no deploy.

## Verification

- `py -3 -m ruff check factorylm_ai/dataset tests/factorylm_ai/test_dataset_v0.py`
- `py -3 -m ruff format --check factorylm_ai/dataset tests/factorylm_ai/test_dataset_v0.py`
- `py -3 -m pyright factorylm_ai/dataset tests/factorylm_ai/test_dataset_v0.py`
- `py -3 -m pytest tests/factorylm_ai -q` -> 471 passed

PR #2881, opened 2026-07-23
fix(factorylm-ai): harden paid Together execution gates

Mikecranesync:
## What

Focused hardening follow-up for merged PRs #2879 and #2880 before any Together spend.

This replaces the nonempty `paid_event_authorization_ref` gate with durable authorization evidence bound to action, dataset manifest, model, spend cap, issuer, expiration, receipt, and unused single-use status. It also removes caller-controlled fine-tune rate overrides, makes dry-run paid intent require the full evidence package, and hardens temporary endpoint cleanup with `inactive_timeout`, immediate endpoint lease persistence, deletion verification, and idempotent orphan cleanup.

## Why

The merged PR4 orchestration was still too easy to fool from a caller: a string authorization could be forged, `usd_per_mtok` could understate cost, budget alone could set `would_create_job`, and endpoint deletion reported success without verifying teardown. This PR closes those false-pass paths while staying pure code.

## Tests run

- `py -3 -m pytest tests\factorylm_ai -q` -> `489 passed in 8.31s`
- `py -3 -m ruff check factorylm_ai tests\factorylm_ai` -> `All checks passed!`
- `py -3 -m ruff format --check factorylm_ai tests\factorylm_ai` -> `68 files already formatted`
- `py -3 -m pyright factorylm_ai\providers\together.py factorylm_ai\pricing.py factorylm_ai\finetune.py tests\factorylm_ai\test_finetune_orchestration.py tests\factorylm_ai\test_flm_ai_providers.py` -> `0 errors, 0 warnings, 0 informations` (existing Pyright config/app-execution-alias notices printed after success)
- `git diff --check` -> exit 0; CRLF warnings only on touched text files
- `py -3 -m pytest -q` -> attempted repo-wide; collection fails on existing monorepo import/conftest conflicts and missing deps (`hypothesis`, `tools.*`, `shared.*`, duplicated `test_cli`/`test_extract` module names), before running this branch's tests

## Safety

No Together call, no upload, no fine-tune job, no endpoint creation, no deployment, and no spend. All live-facing paths are exercised only through fake `httpx.AsyncClient` tests.


Mikecranesync:
## Blocker: the trusted authorization ledger is still self-authorizable

`PaidAuthorizationLedger.record_authorized()` accepts a caller-created `PaidEventAuthorization` and appends it as a trusted `authorized` event. The paid provider then accepts that same object after exact ledger equality and consumes it. There is no signature verification, operator-only intake boundary, trusted external receipt lookup, or separate issuer credential between minting and consuming.

That means a caller able to reach the ledger API can construct an authorization with `issued_by="mike"`, a nonempty authority/receipt ref, the desired request hash and spend cap, call `record_authorized()`, and then pass it into the paid path. The append-only and single-use mechanics prevent replay, but they do not establish that Mike actually approved the first authorization.

Required before merge:

1. Separate authorization issuance from consumption behind a trusted verifier/intake boundary.
2. Verify a signed receipt or trusted operator-controlled record that ordinary paid-path callers cannot mint.
3. Do not expose a general-purpose `record_authorized(caller_supplied_auth)` path to the same trust domain as job creation.
4. Add an adversarial test proving a caller-created authorization cannot become trusted merely by writing it to the ledger.
5. Preserve atomic single-use consumption, request-hash binding, revocation, and fail-closed behavior.

No merge requested. No Together call, upload, endpoint, deployment, authorization consumption, or spend should occur while fixing this.

Mikecranesync:
## 🔴 IMPORTANT — paid authorization trust boundary still forgeable

🔴 IMPORTANT: `factorylm_ai/providers/together.py` exposes `authorization_verifier` on paid public functions, so a caller can pass a fake `PaidAuthorizationVerifier` that returns `trusted=True, consumed=True`; remove caller-controlled verifier injection and construct the verifier inside the provider from trusted operator configuration, with a private monkeypatchable factory only for hermetic tests.

🔴 IMPORTANT: `factorylm_ai/finetune.py::PaidAuthorizationLedger.record_authorized()` promotes an unsigned caller-created `PaidEventAuthorization` into the trusted ledger; require cryptographic authentication of the canonical authorization payload before enrollment/consumption (operator-held key or signed receipt), reject missing/invalid signatures fail-closed, and ensure the paid runtime possesses verification capability but not a general API-callable minting path.

🔴 IMPORTANT: Add adversarial tests proving (1) a caller-created receipt claiming `issued_by="mike"` cannot become trusted by writing it to the ledger, (2) a fake verifier cannot be injected into `create_finetune_job` or the temporary endpoint benchmark, (3) altered signed fields fail verification, and (4) valid signed receipts remain atomic, single-use, revocable, request-bound, and secret-free in emitted evidence.

Keep all network calls fake. Do not upload, train, create an endpoint, deploy, consume a real authorization, or spend money. After the security fix, reconcile with current `main`, preserve the scheduled-monitor changes, choose the next non-conflicting VERSION, and rerun focused pytest, Ruff, format, Pyright, and `git diff --check`.

Mikecranesync:
## Reconciled: 4 red safety tests fixed + merge conflict resolved

Picked this up because it was `CONFLICTING` with 3 red checks and it is the gate standing between the technician-LoRA program and real Together spend.

### Root cause of the 4 failures

`_select_verifier` decided whether to accept a caller-supplied verifier by sniffing **`PYTEST_CURRENT_TEST` + a temp-dir path + the filename `paid-authorizations.jsonl`**. That heuristic was wrong in both directions:

- **Unsound as a boundary** — any caller able to set an env var and drop a well-named file in the temp dir could inject a verifier into a spend path. That is a weaker precondition than the `authorization_verifier=` kwarg it was meant to replace.
- **It swallowed the safety tests** — the four failing tests deliberately point a *genuine* `PaidAuthorizationLedger` at a bad path (`missing.jsonl`, a missing parent, a directory) to prove an unavailable ledger blocks the request **before any HTTP call**. Those paths can't satisfy the heuristic, so the guard rejected them with `PaidAuthorizationRejected` before they ever reached the logic under test.

Worth flagging explicitly: **two of the four pre-exist on `main`** (`test_create_finetune_job_requires_paid_authorization_before_http`, `test_temporary_endpoint_requires_paid_authorization_before_http`) — so this branch had *regressed* them. The other two were added by this PR and never passed.

### Fix

Replaced the heuristic with `allow_ledger_injection_for_tests()` — **off by default, greppable, never called from production**. The type check stays, so a duck-typed always-approve verifier is rejected in *every* configuration. `tests/factorylm_ai/conftest.py` opens the seam via an autouse fixture that restores the prior value per-test, so it can't leak into another suite in a whole-repo run.

### Evidence

| Check | Before | After |
|---|---|---|
| `pytest tests/factorylm_ai` | 528 passed / **4 failed** | **549 passed** / 0 failed |
| `ruff check` | pass | pass |
| `ruff format --check` | pass | pass |
| `pytest tests/test_architecture.py` | — | 9 passed |

Three regression tests lock the boundary: production default rejects even a genuine ledger; the old env+temp-path bypass grants nothing; the seam never admits a duck-typed verifier.

**Mutation-checked** (these tests bite): reducing the guard to a type-only check makes the first two fail — and the request escalates past authorization to the `NetworkDisabledError` gate, which is exactly the progression-toward-spend they exist to catch.

### Merge conflict

`VERSION` only. Resolved to **3.213.1** over main's 3.213.0; renamed the changelog fragment `3.211.2-…` → `3.213.1-…` and noted the seam change in it.

### Safety

No Together call, no upload, no fine-tune job, no endpoint, no deployment, no spend. All live-facing paths exercised only through fake `httpx.AsyncClient`. **Not merging — that's Mike's call.**

PR #2916, opened 2026-07-26
docs(wiki): backfill 13 stranded eval-fixer run reports (2026-07-10 → 07-26)

Mikecranesync:
## Backfill: 13 stranded eval-fixer wiki reports

`wiki/hot.md` on `main` last recorded an eval-fixer run on **2026-06-28**. Thirteen nightly
reports since then (2026-07-10 → 07-26) were committed only to
`fix/precommit-untracked-sigpipe`, which carries a large pile of unrelated accumulated work
and has no open PR — so the reports never reached anyone reading the wiki normally.

This is the `#2759` failure mode (eval-fixer output stranded on an unpushed/unmerged branch)
one layer down: the commits existed, but not on a path to `main`.

### What this PR does

- **Docs only.** One file, `wiki/hot.md`, **+81 lines, 0 deletions** — pure append.
- Cherry-picks *only* the missing `## eval-fixer run` blocks onto main's current file, rather
  than overwriting it. The source branch's copy of `wiki/hot.md` is behind main by 32 lines of
  newer non-eval content; taking it wholesale would have reverted that. It does not.

### Entries added

2026-07-10, 07-11, 07-12, 07-13, 07-14, 07-16, 07-15, 07-17, 07-18, 07-21, 07-22, 07-25, 07-26

(07-15/07-16 are appended in source order; the log is append-only so this is cosmetic.)

### ⚠️ CI caveat for the reviewer

Docs-only PRs in this repo are known to deadlock: the top commit skips `staging-gate`, while
`Version Bump` is a required check (see `docs/versioning.md`). I deliberately **did not** bump
`/VERSION` — that counter is shared with in-flight PRs and unilaterally bumping it from an
automated run seemed worse than flagging it. Bump it or use the documented override as you
prefer.

### Context

Today's run report (27/57, and why the scorecard is confounded by provider timeouts) is on the
rolling tracker: #1876.

🤖 Generated by eval-fixer agent


Mikecranesync:
Superseded by #2942, merged as `caa244e4`.

#2942 backfills the same 13 runs (07-10 → 07-26) plus the 07-27 entry, and its versions of the entries match the source branch verbatim — I diffed `b09b68d2:wiki/hot.md` against both PRs to confirm. This PR's 07-26 entry carried 4 bullets that were **authored here rather than copied from the branch**, so it wasn't a strict superset either way. Disposition of those four:

- 🚨 *"this scorecard does not measure main"* — **kept, and strengthened.** This was the most valuable finding in either PR and I verified it independently: `run-eval-fixer.sh:7` pins `REPO=/Users/charlienode/MIRA` (the shared tree) and `safe-cron-pull.sh` only pulls when that tree is on `main`, so it no-ops otherwise. It's now recorded on `main` in \#2942 as a **mechanism** rather than a behind-by-N count — the counts quoted here (444) and in #2942 (334) disagree with each other and with today's actual 465, because the gap grows every time `main` moves. Root cause filed as #2952; the runner's stranded budget/safe-pull fix is rescued in #2951.
- *"underlying pass rate is sliding"* — **already present** in #2942 as the "Genuine degradation, not just noise" bullet, same 12→20 numbers.
- *"the timeout confound is NEW"* and *"#1632 timeout precedent"* — **dropped deliberately.** Both are analysis of scorecards that we now know measured stale code, so acting on them would compound the error rather than correct it.

Nothing unique is lost. Closing to avoid a conflicting second append to the same region of `wiki/hot.md`.

PR #2971, opened 2026-07-28
fix(ingest): stamp real PDF page in source_page, not the chunk ordinal (#2968)

Mikecranesync:
Write-site half of #2968 (the root cause behind #2910's "PowerFlex 525 · p.1254" citations). Stops **new** mis-stamped rows at the source; existing-row remediation stays tracked in #2968, and the shipped #2910 render guard (#2967) contains them meanwhile.

## Root cause
The offline OEM/document ingest paths stamped `knowledge_entries.source_page` with the **chunk ordinal** instead of the real PDF page. The real page was available all along (`page_num`) — it was just discarded for `source_page`.

## The coupling that made it non-trivial
The shared dedup guard `knowledge_entry_exists()` (in `mira-ingest/db/neon.py`, called by `ingest_manuals.py`, `ingest_gdrive_docs.py`, `ingest_gmail_takeout.py`) matched on `source_page = chunk_index` — so idempotency **relied on the bug**. You can't fix one without the other. Fixed atomically:

| Site | Before | After |
|---|---|---|
| `neon.py` dedup (`knowledge_entry_exists`) | `source_page = :chunk` | `(metadata->>'chunk_index')::int = :chunk` |
| `neon.py::insert_knowledge_entry` | `source_page: chunk_index` | `source_page: page_num` (real page or NULL) |
| `ingest_manuals.py` (own inserts ×2) | `source_page: chunk_idx` | `source_page: chunk.get("page_num")` |

- The new dedup key matches the crawler's already-correct `store.py` `ON CONFLICT (tenant_id, source_url, ((metadata->>'chunk_index')::int))` convention.
- Keying on `source_page` would break idempotency the instant a real page repeats across chunks (many chunks share one page) — the metadata ordinal is the only stable key.
- The chunk ordinal remains in `metadata.chunk_index`. **Existing rows already carry `metadata.chunk_index`**, so re-runs still dedup correctly against them (no duplication).
- Fixing `neon.py::insert_knowledge_entry` fixes the gdrive + gmail paths too — they route through it and already pass `page_num`.

## Evidence (TDD)
- 3 new regression tests in `db/test_knowledge_entries.py` (written first, red before the fix): `source_page == page_num` (not the ordinal), `source_page IS NULL` when page unknown, and dedup keys on `metadata->>'chunk_index'`.
- **19 passed + 3 skipped** in `db/` (16 existing + 3 new). ruff clean.
- Scope: **offline/manual ingest scripts — not in the live crawler daemon** (verified against `job_registry.py`). The live `mira-crawler` pipeline was already correct.

## Not in scope (tracked in #2968)
Backfilling/de-duplicating the ~61.8k existing mis-stamped rows + the 158×-duplicated docs — that's the gated prod-data remediation program.


Mikecranesync:
Superseded by #2976 (merged). Both PRs fix the same root cause (chunk ordinal in source_page). The unit tests added here are the only delta — those can be a lightweight follow-up if needed, but the core write-site fix and dedup key fix are now on main.

PR #3018, opened 2026-07-30
feat(ignition): one canonical Gateway live-snapshot adapter — both PLC transports render the same typed reading

Mikecranesync:
The PLC-laptop / Ignition contribution to MIRA unification. **Merge-ready**: 27/27 CI green, rebased on `main`, live-gateway shape evidence recorded below.

## The defect

The Gateway emitted **two incompatible shapes for the same tag read**:

| | CHAT path (`api/chat/doPost.py`) | STREAM path (`gateway-scripts/tag-stream.py`) |
|---|---|---|
| `value` | `str(qv.value)` — **always a string** | real Python type |
| `quality` | raw Ignition string (`Good_Unspecified`) | **banded** `good\|bad\|stale\|uncertain` |
| allowlist | `is_allowed_tag()` inside `except ImportError: pass` → **fail-open** | **fail-closed** |
| timestamp | `str(qv.timestamp)` | `ts` |

Same gateway, same instant, two truths: `value` was sometimes `48.5` and sometimes `"48.5"`; `quality` sometimes `good`, sometimes `Good_Unspecified`.

The **fail-open allowlist** was the sharpest edge — if that one import failed, a **non-allowlisted tag could reach the cloud**.

## The fix — one adapter, two renderings

`ignition/webdev/FactoryLM/api/tags/live_snapshot.py` is now **the** Gateway live-snapshot adapter:

```
      system.tag.browseTags / readBlocking      (injected, read-only)
                      │
        live_snapshot.read_tag_readings()
                      │
         collector.build_reading()              ONE typed reading
                      │
        collector.filter_allowlisted()          fail-closed
                      │
        ┌─────────────┴─────────────┐
 STREAM  collector.build_payload()   snapshot_from_readings()  CHAT
         list → mira-relay           dict → ignition_chat.py
```

`doPost.py` no longer reads tags itself.

**This is not** a new ingestion system, transport, normalizer, parser, relay, database writer, or context schema — it's a *renderer* over readings that already existed. `materialized_evidence/` untouched. `mira_diagnose`/A0–A12 unchanged.

## Behaviour that is deliberately load-bearing

- **Fail-closed:** no resolvable allowlist ⇒ **empty** snapshot, never unfiltered. `doPost.py` logs that at **error** level so it can't be mistaken for "this asset has no tags." A degraded answer beats a broken tag contract.
- **Never fabricate:** a ragged/short `readBlocking` result **skips** the tag rather than padding it. A made-up number in front of a technician is the worst outcome here.
- **Degrade, don't raise:** Gateway I/O failure ⇒ empty snapshot, so the turn stays answerable from documentation.

## Read-only, asserted mechanically

The adapter imports **no** Ignition runtime and reaches a tag only through injected callables — there is no code path that can write a tag, send a command, or touch a fieldbus (`.claude/rules/fieldbus-readonly.md`). Two tests assert it on the source rather than trusting review.

## Two defects in my own guards, found by mutation-testing them

Both looked fine, and this is the part most worth reviewing:

1. **v1 matched raw file text** → fired on the *comments describing* the removed behaviour. A guard reading prose, not code.
2. **v2 stripped comment/string tokens and re-joined with `" ".join(...)`** → made every **multi-token** assertion vacuous. `system.tag.write` tokenizes to `system . tag . write`, so a mutation that reintroduced `str(qv.value)` **PASSED**. Only single-token names like `writeBlocking` were genuinely checked.

The stripper now blanks comment/string spans **in place**, preserving layout so dotted paths survive verbatim. `test_code_only_preserves_multi_token_syntax` pins it. Re-running the mutations confirms both are caught — including a dotted `system.opc.write` the earlier version silently missed:

```
MUTATION reintroduce str(qv.value) in doPost  → FAILED test_chat_handler_no_longer_reads_tags_inline
MUTATION add system.opc.write to adapter      → FAILED test_adapter_contains_no_write_or_fieldbus_path
restored                                       → 18 passed
```

## Verification — all offline, injected/mocked

- `tests/ignition/` **92 passed** (18 new). No Gateway, PLC, network, Neon, Doppler, or secrets.
- `tests/test_architecture.py` **13 passed** — the one-pipeline contract still holds.
- Jython-2.7-safe: AST-checked, no f-strings or annotations, so it imports identically in the Gateway and under pytest.
- `ruff check` clean on the new test.
- **`test_chat_and_stream_paths_agree_on_every_reading`** asserts `value` / `value_type` / `quality` field-by-field across both renderings — the unification claim, tested rather than asserted.

## Boundaries honoured

Untouched: CCW projects, ladder/ST logic, bench-control scripts, `plc/live_monitor.py`, production secrets, `materialized_evidence/`. No writes, no PLC commands, no gateway deployment.

## Docs

- **`NORTH_STAR.md`** — new doctrine: *one evidence shape per producer.* "Every adapter renders the same approved-context answer" is only true if every adapter is handed the same *evidence*; so a producer has exactly one adapter and transports are renderings over it.
- **`docs/integrations/ignition-tag-collector.md`** — the flow diagram, the old-vs-new table, deployment note (all four modules ship together in the script library), and the rules the adapter must keep.

## Reviewer notes / known gaps

- `doPost.py`'s injected `_browse`/`_read` wrappers are the only place `system.tag.*` is referenced; the guard allows **at most one** occurrence of each so a second inline read can't creep back.
- Not addressed here (out of scope, flagged not fixed): `doPost.py` will still forward a turn with an **empty `asset_id`**, which `.claude/rules/direct-connection-uns-certified.md` says a direct-connection surface should *reject* rather than downgrade. That's a chat-contract question, not a snapshot-shape one.
- No live Gateway verification — explicitly out of scope. First real deployment should confirm the error-level log fires when `approved_tags.json` is missing.

---

# UPDATE (2026-07-31, commits `3aa6ce4cf` + `df619a55b`): the deployment contract — this adapter could not have run on a Gateway

Everything above is correct and unchanged. But the deployment path never produced a working WebDev endpoint, and every way it failed was silent — so the adapter, however right, could not have taken effect. Established against the live 8.3.4 bench gateway, then encoded here.

## The three deployment defects

1. **A WebDev endpoint is a THREE-file resource** — `resource.json` (its `files` array *is* the data-key list), `config.json` (`resource-type` + one **kebab-case** block per method), and a def-first `<method>.py`. `deploy_ignition.ps1` shipped only the `.py`: `HTTP 500 No data found for resource`.
2. **The repo's nine handlers are comment-first Python modules.** `PythonResource.maybeRemoveDef` strips only a *leading* `def`, so a module-style body is not rejected — it returns **HTTP 200 with an EMPTY body**, no log line. It looks deployed and healthy. This is why the defect survived.
3. **Helpers beside a handler are not importable.** Inside a resource, `sys.path` is only the gateway's pylib dirs and `__file__` is undefined. Helpers must be **project script-library** modules, imported as flat top-level names.

## What this update adds

- **`ignition/tools/webdev_build.py`** (new) — deployment is now a *conversion*, not a copy: emits both manifests + a def-first body per handler (nesting headers/imports/helper `def`s inside the function, incl. the two *after* the handler in `api/connect/doPost.py`), and **refuses** to emit anything that isn't exactly one module-level `def <method>(request, session)`. Python, not PowerShell, so it's unit-testable offline.
- **`deploy_ignition.ps1`** — calls the converter; deploys `signing`/`allowlist`/`collector`/`gateway_live_snapshot` to the script library; refuses to overwrite a live project without `-Force`; prints project-qualified URLs (`/system/webdev/<PROJECT>/FactoryLM/...`). Also: the script **did not parse on origin/main** — 22 UTF-8 em-dashes, no BOM, 7 parser errors under PowerShell 5.1's Windows-1252 fallback. Now pure ASCII, 0 errors.
- **One authoritative allowlist location** — deploy wrote where the loader never looked (`ConveyorMIRA` installed vs `…/projects/factorylm/…` searched). Both now key off `allowlist.RUNTIME_ALLOWLIST_RELPATH` under the Ignition *data* dir; a test pins the agreement.
- **Explicit-but-unreadable `MIRA_ALLOWLIST_PATH` now fails closed** instead of silently selecting a different default allowlist. ⚠️ Deliberate contract change: `test_env_var_nonexistent_falls_through` renamed + inverted, reasoning in its docstring.
- **`tag-stream.py`** import order flipped to prefer the *proven* flat script-library route; the third unguarded `__file__` site (`api/diagnose/doGet.py`) guarded, with an AST test pinning the class repo-wide.
- **33 new tests** in `tests/regime7_ignition/test_webdev_deploy_contract.py`; **231 pass** across both Ignition suites; 27/27 CI checks green.

## Live gateway evidence (Ignition 8.3.4, 2026-07-31)

The "No live Gateway verification" note above is now superseded **for the resource shape**. Three resources, one restart, `GET` each:

| body shape | result |
|---|---|
| `def` on line 1, 4-space body — *what the converter emits* | **HTTP 200** `{"shape":"a"}` |
| comment header, then `def` at col 0 — *the raw repo shape* | **HTTP 200, len=0** (the silent failure, reproduced) |
| `def` first + module-level helper *after* it | **HTTP 200** `{"shape":"c","helper":"helper-called-with-ok"}` |

Plus, separately probed: `from signing import build_headers` resolves from the **script library** inside a WebDev resource (`OK: <function build_headers …>`), and a sibling file does **not** (`ImportError`). Probe resources deleted afterwards (verified 404).

**Still unproven, stated plainly:** the nine *converted* handlers end-to-end on a gateway. Post-merge step: `deploy_ignition.ps1 -ProjectName MiraDeployTest`, hit `/system/webdev/MiraDeployTest/FactoryLM/api/status` + one POST endpoint, delete the scratch project.

**Unchanged on purpose:** the canonical adapter, fail-closed allowlist logic, typed values, quality bands, one-read path, no PLC/OPC writes.

---

# UPDATE 2 (2026-08-01, commits `afed363aa` + `6b41f0c99`): adversarial review — five findings, all verified real, all fixed

Review: `docs/handoffs/2026-08-01-pr3018-adversarial-review-for-claude.md`. Every finding was reproduced before it was fixed.

| # | finding | verification | fix |
|---|---|---|---|
| 1 | Merge conflicts with `origin/main` | reproduced (`VERSION` + `docs/CHANGELOG.md`) | rebased onto `v3.237.0`; changelog rebuilt as main's file + one consolidated `v3.237.1` entry; merge dry-run now `Already up to date` |
| 2 | Activation writes `RELAY_URL`, stream reads only `INGEST_URL` | confirmed in both files — a freshly activated gateway streamed to the hardcoded default | stream reads `INGEST_URL` (manual override, wins) → `RELAY_URL` → default; activation deliberately never writes `INGEST_URL`; pair + precedence pinned on comment-stripped code |
| 3 | "activated" reported with nothing persisted | confirmed — `_write_config()` no-ops when no properties file exists | updates in place, **creates** the file under the first existing Ignition data dir, returns True only on disk write; failure → explicit **500** carrying the tenant id; 3 new handler tests incl. the clean-gateway case |
| 4 | `-Force` copy nests into the existing dir and keeps stale files | **reproduced live** (`dst\src` + stale file survived) | backup → `Remove-Item` destination → copy; ordering pinned by test |
| 5 | Non-ASCII UTF-8 in deployed Jython artifacts, no encoding declaration | confirmed: all 13 artifacts non-ASCII; the only module proven running on the gateway is pure ASCII, so undeclared UTF-8 is an unproven risk | PEP 263 cookie stamped on every artifact — line 1 for script-library modules, first line *inside* the def (file line 2) for handlers, the only placement compatible with the def-first rule; harmless if Ignition compiles decoded strings, load-bearing if bytes |

Verification block from the review, run: `tests/ignition` + `tests/regime7_ignition` + `tests/test_architecture.py` → **242 passed** (11 new this round, 44 across the PR); ruff clean on all touched files; `git diff --check` clean; `deploy_ignition.ps1` parses with 0 errors and is pure ASCII; GitHub reports **`MERGEABLE / CLEAN`**, 27/27 checks green.

Not done, per the review's constraints: no live-gateway deployment (needs Mike's explicit approval); no PLC logic, CCW, fieldbus write paths, or production secrets touched.

https://claude.ai/code/session_01Dgej4vU7mMnYTmU6QXsH3n


Mikecranesync:
## Review of the guards, per the request in the PR body — plus a fix pushed as `5742ae963`

The PR asked its reviewer to *"re-check the guard logic rather than the feature."* Did that by mutation testing. **23 mutations, 16 red, 7 green.** The seven greens matter, and one thing the guards can't see would have made this feature dead on arrival.

### 🔴 Production-fatal: `__file__` is undefined in an Ignition script resource

`collector.py:35-43` — in this PR's own dependency — documents exactly this trap and guards it with `except NameError`. The new adapter import in `doPost.py` did the same `__file__` path dance **unguarded**, inside a broad `except Exception` that logs at **warn**:

```
_api_dir = _osp.dirname(_osp.abspath(__file__))     # NameError on a Gateway
...
except Exception as e:
    logger.warn("Live snapshot unavailable ...")     # swallowed
```

On a real Gateway that NameError is swallowed and **every chat turn ships an empty snapshot** — indistinguishable from "this asset has no tags". No CPython test can catch it, because `__file__` always exists under pytest, and `tests/regime7_ignition/` executes `doPost` under CPython too. Now guarded (a flat sibling import is the Gateway layout anyway), and a missing adapter is its own `except ImportError` at **error** level, since it's a deployment fault, not a transient read failure.

### 🔴 Four guards could not fail on the defect they name

| Mutation | Was | Why |
|---|---|---|
| `system.tag.write([asset_id],[0])` added to **`doPost.py`** | 🟢 green | the read-only gate scanned only `gateway_live_snapshot.py` — the one file that is read-only **by construction** (imports no Ignition runtime, tags only via injected callables). `doPost.py` is the file with real `system.*` access. |
| the whole original inline stringifying snapshot re-added to `doPost.py` with `unicode()` instead of `str()`, reusing the existing `_browse`/`_read` wrappers | 🟢 green | the guard pinned a **spelling**: `unicode(` isn't `str(qv.value)`, no `is_allowed_tag`, and no new `browseTags`/`readBlocking` token so both counts stayed ≤ 1. |
| fail-open reintroduced inside `if allowlist is None:` | 🟢 green | that branch is the **only one production takes** (`doPost.py` passes no allowlist) and had **zero** coverage — every fail-closed test passes an explicit `allowlist=`. `--cov` confirms `gateway_live_snapshot.py:185-188` and `collector.py:110-113` missing. |
| `if not asset_id: return {}, stats` deleted | 🟢 green | `test_missing_asset_id_yields_no_snapshot` used `make_io({})`, whose `browse_fn` returns `[]` for *any* folder — it asserted over its own fixture. |

Plus: the two-transport parity test looped `("value","value_type","quality")` — omitting **`ts`**, so the renderings could disagree on observation time undetected.

### What `5742ae963` changes

- OT-write sweep now covers **every `.py` under `ignition/webdev/**`**, not one hand-picked module. DB writes stay allowed deliberately (`runPrepUpdate` is legitimate — chat history, alarm ack); `gateway-scripts/` memory-tag alert writes (`tag-change-fsm-monitor.py:68`, `timer-stuck-state.py:238`) are pre-existing and **explicitly out of scope**, not silently swept in.
- The defect **class**, asserted structurally: `filtered_snapshot[` is never assigned into, and no `str(`/`unicode(` is applied to any `.value`.
- Production allowlist branch covered both ways; assetless turn proves `browse` was never **called**; `ts` in the parity loop; `except NameError` / `except ImportError` pinned by a test.
- Stale header ref `tests/ignition/test_live_snapshot.py` → `test_gateway_live_snapshot.py` — the exact basename collision the rename commit existed to remove.

**Evidence:** 6/6 mutations RED, each caught by its own guard, byte-exact restore (md5), green after. `tests/ignition/` **97 passed**; `tests/regime7_ignition/` **93 passed** (that suite actually executes `doPost.py`). `gitleaks` clean; `actionlint` clean on `ci.yml`.

### ⚠️ The PR's CI premise is false — and the true version is more useful

> "`tests/ignition/` was run by NO workflow"

Not so. `test-eval-offline` already swept it — `pytest tests/ -v -n auto -m "not network and not slow"`, and no `--ignore` covers `tests/ignition/`. **184** matching lines in this PR's own run (`30516072173`), including the new file. `tests/regime7_ignition/` was running there too.

What was true: `test-eval-offline` is **not in `ci-gate.needs`** ("heavy quality gates … promote once reliable"), so the suite ran on every code PR and **blocked nothing**. Adding it to the gated `test-unit` job is the right fix for a different reason. Comment and CHANGELOG corrected.

And one level up, which no workflow file can fix: **`CI Gate` is not itself a required status check on `main`** — required is `staging-gate`, `Version Bump Check`, `Hub E2E`, `mira-web pack tests`. `test-unit` / `architecture-check` / `lint` / `sast-*` are all advisory today. Generalised as **Contract 8** in #3021, which also flags that the pending removal of `Version Bump Check` should **add `CI Gate`** in the same visit — otherwise required checks drop to three and test enforcement stays at zero.

### Flagged, not fixed

- `resolve_allowlist_path()` honours `MIRA_ALLOWLIST_PATH` only `if os.path.isfile(override)` and otherwise **falls through to the default search paths** — a typo'd override silently substitutes a *different* allowlist (the installed-Gateway file, 58 tags on the bench laptop) instead of failing closed.
- The write sweep is a token scan: a write via `getattr(system, "tag")` indirection, or through an injected callable, is **not** detected — both confirmed to slip through. That limit is written into the test file rather than implied away.
- `collect_live_snapshot`'s "returns empty rather than raising" contract isn't quite true: a non-`set` allowlist propagates `AllowlistError`. Production is shielded only because `doPost.py` wraps the block.
- Still open from the original body: `doPost.py` forwards an empty `asset_id`. Checked end to end — the contract **is** met, cloud-side: `ignition_chat.py:664-674` returns `422 {"error":"uns_required"}` for an asset-specific turn with no identifier, tested in `mira-pipeline/tests/test_ignition_chat_direct_connection.py`. Two docs still claim it's unimplemented (`docs/runbooks/activate-ignition-ask-mira.md:39`, `docs/demos/_audit/ignition-audit.md:198`) — stale, worth a follow-up so nobody "fixes" a non-bug. Note `_is_asset_specific` fails **open** by design; that's a policy call, not a defect.

Unverified, and it needs the bench Gateway: whether a WebDev resource defines `__file__` at all (`api/tags/doGet.py:42,69` already relies on it), and that the fail-closed `approved_tags.json` error log actually fires there.

https://claude.ai/code/session_01Dgej4vU7mMnYTmU6QXsH3n

Mikecranesync:
## Corrections to my review above, and a second fix — `23764ad66`

I ran three adversarial skeptics against my own hardening commit. They refuted it. Corrections first, because two things I wrote above were wrong.

### 🔴 The `__file__` fix was HALF done — and the half I missed is worse

`doPost.py` does the path dance **twice**. I guarded the adapter import and missed the **signing** one ~60 lines below, which sits **outside any `try`**, at function-body level. `doPost` has no outer `try` (structure: `try` at 99 → `except ImportError` → `except Exception` → then bare code). So on an Ignition script resource that `NameError` is **uncaught**: the endpoint returns **HTTP 500 on every chat turn** — not a degraded answer, no answer.

My own comment made that harder to find, not easier: it described the consequence of an unguarded `__file__` as "warn + empty snapshot", which is true *inside* the try and false 60 lines down. A future reader would have read it and concluded the class was handled. Both sites are guarded now, and the comment says the consequence depends on where the dereference sits.

The guard is now **structural over the AST** — every `__file__` Name node must be lexically inside a `try` whose handlers include `NameError` — applied to `doPost.py`, `gateway_live_snapshot.py` and `collector.py`. My first version was `assert "except NameError" in code`: a bare substring, unbound to the expression it protects, and a skeptic satisfied it with a **decoy handler** while leaving the real dance bare. That is the *third* generation of one mistake in this file — matching text that describes the invariant instead of the code that carries it.

### 🔴 Three of my four "closed" holes were still open

| Evasion (all were green) | Fix |
|---|---|
| `system.tag.configure(...)` and `system.tag.editTag(...)` — both mutate tags, neither in my blocklist, and any future `system.tag.<anything>` would slip too | the sweep is now **default-deny** over the tag/OPC namespaces: only `_TAG_READ_MEMBERS` / `_OPC_READ_MEMBERS` may be touched from the WebDev surface, so an unanticipated API fails closed. Enumerating writes is a losing game. |
| the whole original defect re-added as a **differently-named** dict (`_inline = {}`, `"%s" % qv.value`, raw quality, fail-open) then merged — `filtered_snapshot[` never appeared | a second read path cannot avoid the injected wrappers, so `_browse` and `_read` must each be referenced **exactly twice** (definition + the single `collect_live_snapshot` call) |
| decoy `except NameError` (above) | AST, per above |

Replayed after the fix: all five evasions **RED**, each caught by its own guard, byte-exact restore (md5), green after. `tests/ignition/` **99 passed**; with `tests/regime7_ignition/` **192 passed**.

### Corrections to the claims in my review

- **`5742ae963` is a dangling sha** — I force-pushed the rebase after posting. Current tip: `23764ad66`. Content impact nil, but a reader following that sha lands off-branch.
- **"the contract *is* met" for empty `asset_id` was misleading.** The code claim is true (`ignition_chat.py:664-674` returns `422 uns_required`, `test_ignition_chat_direct_connection.py` asserts it, 14 passed locally). But **that test is executed by no workflow** — `mira-pipeline/tests` is one of the exact directories I counted as unreached in #3021. Citing it as closure, in a comment arguing that a check which can't fail a merge isn't a guard, was self-contradictory. The contract is met *in the code*, unguarded *in CI*.
- **184 vs 74:** both true, different populations, stated in adjacent clauses. 184 log lines = 92 distinct node ids on the PR branch; 74 is `origin/main`'s baseline before this PR adds a file.

### Still open — flagged, not fixed

- The fail-open `except` at `gateway_live_snapshot.py:187-188` is **still uncovered** per `coverage report`, despite the two new allowlist tests.
- `system.device.*` and test steps invoked via `uses:` are outside these guards.
- **`doPost.py` still forwards an empty `asset_id`.** `.claude/rules/direct-connection-uns-certified.md` § "What the audit must catch" names this first. The cloud does return `422 uns_required` — but `doPost.py` renders it as a generic `"MIRA Cloud returned error", http_status: 422`, so a **contract violation looks like an outage**. It's ~5 lines mirroring the existing empty-`query` guard, and it's now been flagged three times (handoff §3.2, the PR body, my review). Left alone only because rejecting *all* assetless turns would break the rule's own educational carve-out, which the cloud handles by classifying intent — so the right fix is rendering the 422 honestly, not rejecting locally. Worth its own PR.
- **`__file__` on a WebDev resource is still unverified.** `api/tags/doGet.py:42,69` already relies on it, so it may well be defined there — but nothing here proves it either way, and the bench Gateway is the only thing that can. That promotes "confirm on first deployment" from nice-to-have to **blocking**: the 500 above is exactly what a live check would have caught, and no CPython test ever will.

https://claude.ai/code/session_01Dgej4vU7mMnYTmU6QXsH3n

Mikecranesync:
## Adversarial review (2026-08-01) — all five findings verified real, all fixed

Review doc: `docs/handoffs/2026-08-01-pr3018-adversarial-review-for-claude.md` · Response doc: `docs/handoffs/2026-08-01-pr3018-adversarial-review-fixes.md`
Fix commits: `afed363aa` (findings 2–5) + `6b41f0c99` (finding 1 / changelog consolidation). Each finding was **reproduced before it was fixed**.

| # | finding | verified | fix |
|---|---|---|---|
| 1 | Merge conflicts with `origin/main` | reproduced (`VERSION` + `docs/CHANGELOG.md`) | rebased onto `v3.237.0`; changelog rebuilt as main's file + one consolidated `v3.237.1` entry; merge dry-run now `Already up to date` |
| 2 | Activation writes `RELAY_URL`, stream reads only `INGEST_URL` | confirmed in both files | `tag-stream.py` now reads `INGEST_URL` (manual override, **wins**) → `RELAY_URL` → default; activation deliberately never writes `INGEST_URL`; pair + precedence pinned on comment-stripped code (`TestActivationStreamConfigContract`) |
| 3 | `"activated"` returned with nothing persisted | confirmed — `_write_config()` no-ops with no properties file | now creates the file under the first existing Ignition data dir, returns `True` only on disk write; persistence failure → explicit **500** carrying the tenant id (`TestConnectActivationPersistence`, 3 tests incl. clean-gateway) |
| 4 | `-Force` copy nests into existing dir, keeps stale files | **reproduced live** (`dst\src` + stale file survived) | backup → `Remove-Item` destination → copy; ordering pinned by test |
| 5 | Non-ASCII UTF-8 in deployed Jython artifacts, no encoding declaration | confirmed: all 13 artifacts non-ASCII; the only module proven running on the gateway is pure ASCII | PEP 263 cookie on every emitted artifact — line 1 for script-library modules, first line *inside* the def (file line 2) for handlers, the only placement compatible with the def-first rule (`TestEncodingSafety`) |

**Verification block from the review, executed:** `tests/ignition` + `tests/regime7_ignition` + `tests/test_architecture.py` → **242 passed** (11 new); ruff clean on all touched files; `git diff --check` clean; `deploy_ignition.ps1` parses (0 errors, pure ASCII); GitHub reports **`MERGEABLE / CLEAN`**, 27/27 checks green at head `6b41f0c99`.

**Constraints honoured:** not merged (Mike's call); no live-gateway deployment this round; PLC logic / CCW / fieldbus write paths / production secrets untouched; the adapter's fail-closed, read-only behaviour unchanged.

**Honest residual:** the nine *converted* handlers end-to-end on a real gateway remain unproven offline (CPython always defines `__file__`). Post-merge, with explicit approval: `deploy_ignition.ps1 -ProjectName MiraDeployTest` → hit `/system/webdev/MiraDeployTest/FactoryLM/api/status` + one POST endpoint → delete the scratch project. Since every deployed artifact is non-ASCII, that same run also settles whether the PEP 263 cookie was load-bearing.

Mikecranesync:
# PR #3018 Adversarial Review Handoff

PR: https://github.com/Mikecranesync/MIRA/pull/3018

Verdict: not ready to merge.

## Fix These Before Merge

1. Resolve the current merge conflicts with latest `origin/main`.
   - Conflicts reproduced locally in `VERSION` and `docs/CHANGELOG.md`.
   - Rebase or merge main, keep the correct new version/changelog entry, then rerun CI.

2. Fix the activation-to-stream config mismatch.
   - `ignition/webdev/FactoryLM/api/connect/doPost.py` writes `RELAY_URL`.
   - `ignition/gateway-scripts/tag-stream.py` reads only `INGEST_URL`.
   - A freshly activated gateway can ignore the relay URL it just received.
   - Make the stream read canonical `RELAY_URL` or have activation also persist the exact key the stream uses. Add a regression test.

3. Do not report activation success unless config was actually persisted.
   - `_write_config()` only writes when an existing `factorylm.properties` exists.
   - If no file exists, it logs a warning but `doPost()` still returns `"activated"`.
   - Either deploy/create the properties file first or make `_write_config()` fail loudly and return an activation error. Add a test for clean-gateway/no-properties-file.

4. Fix `deploy_ignition.ps1 -Force` overwrite behavior.
   - Current code backs up the existing project, then runs:
     `Copy-Item -Path $ProjectSrc -Destination $ProjectDst -Recurse -Force`
   - In PowerShell, copying a directory to an existing directory nests it as `$ProjectDst\<source-name>` and leaves stale files in place.
   - After backup, remove the existing destination or copy the contents (`$ProjectSrc\*`) into a clean destination. Add a test or script guard.

5. Prove or eliminate Jython source-encoding risk.
   - Generated Gateway/WebDev artifacts contain UTF-8 non-ASCII bytes with no source encoding declaration.
   - CPython tests read UTF-8 explicitly, so they do not prove Ignition/Jython 2.7 parses the deployed artifacts.
   - Either make deployable Jython-targeted sources ASCII-only, add `# -*- coding: utf-8 -*-` if Ignition/Jython accepts it, or run/record a live Gateway parse probe for the exact generated artifacts.

## Verification Required

Run at minimum:

```powershell
git fetch origin main
git merge --no-commit --no-ff origin/main
git merge --abort
py -3 -m pytest tests\ignition tests\regime7_ignition\test_webdev_deploy_contract.py tests\test_architecture.py -q
py -3 -m ruff check ignition\tools ignition\webdev tests\ignition tests\regime7_ignition
git diff --check
```

Also verify GitHub reports `mergeable=MERGEABLE` and `mergeStateStatus=CLEAN`.

## Constraints

- Do not merge until the above blockers are fixed.
- Do not deploy to a live Gateway unless Mike explicitly approves it.
- Do not touch PLC logic, CCW projects, fieldbus write paths, or production secrets.
- Keep this as one focused hardening update to PR #3018.

Mikecranesync:
## Compliance check vs the re-posted review (2026-08-01 13:41) — all five blockers fixed at HEAD `6b41f0c99`, re-verified just now

The 13:41 comment is the same five findings as `docs/handoffs/2026-08-01-pr3018-adversarial-review-for-claude.md`; the fixes landed in `afed363aa` + `6b41f0c99` (detail in the 12:30 comment above). Re-running the required verification block **at current HEAD, against current `origin/main`**:

```
git rev-list --count HEAD..origin/main        -> 0
git merge --no-commit --no-ff origin/main     -> Already up to date
pytest tests/ignition tests/regime7_ignition/test_webdev_deploy_contract.py tests/test_architecture.py -> 159 passed
pytest tests/regime7_ignition tests/ignition (full) -> 242 passed
ruff check (touched files)                    -> clean  (8 pre-existing findings remain in
                                                 test_diagnose_endpoint.py / test_gateway_scripts.py — untouched by this PR)
git diff --check                              -> clean
GitHub                                        -> MERGEABLE / CLEAN, 27/27 checks green
```

Blocker-by-blocker at HEAD: **(1)** 0 behind main, merge dry-run clean · **(2)** `tag-stream.py:190-192` reads `INGEST_URL` → `RELAY_URL` → default, pinned by `TestActivationStreamConfigContract` · **(3)** `_write_config()` creates-or-fails-loudly, `doPost()` 500s on persistence failure, pinned by `TestConnectActivationPersistence` (incl. the clean-gateway case) · **(4)** backup → `Remove-Item $ProjectDst` → copy, ordering pinned by `test_force_overwrite_removes_stale_destination_before_copy` · **(5)** PEP 263 cookie on all 13 artifacts, pinned by `TestEncodingSafety`.

---

### Also answering the automated pipeline review (11:37)

- 🔴 *"`except ImportError: pass` in `api/chat/doPost.py` could mask security issues"* — **stale finding**: at HEAD the only `except ImportError: pass` text in that file is inside a **comment describing the removed fail-open behaviour** (line 115). The live handler logs the ImportError at **ERROR** with a deployment pointer (`doPost.py:176-185`), and the snapshot stays empty (fail-closed). The reviewer matched prose, not code — the same defect class this PR mutation-tested its own guards for.
- 🔴 *"unclear what happens if the `-Force` overwrite fails"* — `$ErrorActionPreference = "Stop"` (line 18) makes any `Copy-Item`/`Remove-Item` failure terminate the script immediately. Ordering is backup → remove → copy, so a failure at any step leaves the timestamped backup intact and never half-overwrites: backup fails → nothing removed; remove fails → original + backup both present; copy fails → backup present, script dead before the converter runs.
- 🟡 *"assumes `build_reading()` is consistent across both paths"* — not assumed: `test_chat_and_stream_paths_agree_on_every_reading` (`tests/ignition/test_gateway_live_snapshot.py:210`) asserts `value`/`value_type`/`quality` field-by-field across both renderings.
- 🟡 *input validation* — the concrete boundaries validate: connect rejects a missing `code` (400) and a malformed activation response (502); chat fails closed (503) without an HMAC key. Broader schema validation is out of this PR's scope.
- 🔵 *rename `INGEST_URL`/`RELAY_URL`* — **declined deliberately**: these are config keys living in fielded `factorylm.properties` files; renaming deployed keys is precisely the two-names-for-one-thing defect class this PR spent three fixes removing. The precedence comment at `tag-stream.py:184-189` documents what each name means instead.

No code changes were needed for the automated round — two findings were already fixed at the flagged locations, one was already tested, one is documented behaviour, one is declined with rationale.

PR #3032, opened 2026-07-30
feat(engine): WS1 — the context contract becomes a live consumer on the serving path (ADR-0033 G1/G6)

Mikecranesync:
## Why

`materialized_evidence/context_contract.py` and its nine adapters shipped in Phase 3 and then **sat unused**. Verified on `main`: `evidence_from_prior_decisions()` had **zero production call sites** — referenced only from `tests/test_context_contract.py`. That gap is not cosmetic; it is the critical path:

```
WS1 adoption  →  eval slice 13 buildable  →  slice manifest freezable  →  training authorization has an acceptance bar
```

`docs/zta/technician-unified/eval-manifest.md` records slice 13 as *"blocked on runtime adoption"*. This PR closes that block for **one** evidence family on **one** serving path, end to end: recall → typed context → validated → prompt → audit row.

The serving path is `Supervisor` (`mira-bots/shared/engine.py`), deliberately — it is shared by `mira-pipeline` (the VPS chat path), Telegram and Slack. Wiring an Ignition-only or PrintSense-only entry would have recreated exactly the forked context lane this program exists to remove.

## What ships

| File | Role |
|---|---|
| `mira-bots/shared/prior_decisions.py` | **The first-ever reader of `decision_traces`** (the Phase-1 inventory found two writers and zero readers) |
| `mira-bots/shared/technician_context.py` | Turn → `TechnicianContext`, validated, + the manifest the audit row stores |
| `mira-bots/shared/engine.py` | `_build_prior_decisions_context()` joins the five existing enrichment blocks; manifest rides the per-turn result to the trace |
| `mira-bots/shared/decision_trace.py` | Stores the manifest **verbatim** — never re-derived |
| `mira-hub/db/migrations/071_…sql` | Additive `context_manifest JSONB` + `context_manifest_sha256 TEXT` |
| `mira-pipeline/Dockerfile` | Ships `materialized_evidence/` — **it did not** |

## Four decisions worth reviewing

**1. The prompt gets only the PRIOR_DECISION projection — not the full contract block.**
`rag_worker._build_prompt_with_chunks` already renders retrieved chunks as `--- RETRIEVED REFERENCE DOCUMENTS ---`. Rendering `to_prompt_block()` whole on top would present the *same* chunk twice under *two* trust labels. So the prompt gets exactly the evidence family that is genuinely new. **Consequence, stated plainly: G6 ("audit row = prompt row") holds today for contract-sourced evidence, not yet for retrieval.** Unifying retrieval's prompt rendering onto the contract is the next slice; this PR does not claim it.

**2. Flag-gated `MIRA_CONTEXT_CONTRACT`, default OFF — and that still closes the block.**
What slice 13 needs is that the flag-**on** path runs the *real* serving route, not that it is on by default. Default-off protects `tests/golden_*.csv` and the eval regimes from an unmeasured prompt-shape change. The flag is read **per call**, not captured at import, so the slice can run one process both ways via `monkeypatch.setenv` — there is a test asserting exactly that. Precedent: `ENFORCE_PLAN_QUOTA`, `ENFORCE_ASSET_AGENT_GATE`, `MIRA_LOCAL_TRACE`.
**Promotion trigger (declared, not left implicit):** flip to on-by-default once slice 13 exists and the regimes show no regression against the flag-off baseline.

**3. Fail-open, but never silent (ADR-0033 requirement 6).**
`decision_trace.py` is fail-open because the write is fire-and-forget — silence there is fine. This **read** happens *before* the answer, so a failure changes what MIRA knows. An attempted-and-failed lookup therefore returns an explicit `prior_decisions_unavailable`, which becomes a `TechnicianContext.unknowns` entry, visible **in the prompt and in the audit row**. "No prior history" and "we could not look" must not be indistinguishable. An *unconfigured* store is deliberately **not** an error — nothing was attempted.

**4. A new column, not an overloaded one.**
The alternative was stuffing the manifest into the existing `kg_evidence`/`tag_evidence` JSONB. That is the "don't fork the contract" smell in reverse — it hides a first-class concept inside a field that means something else. Migration 071 is additive, idempotent, drops no policy or index (so none of the rule-§4 ALTER ordering applies), and leaves `tenant_id` untouched. NULL reads correctly as *"predates the contract, or the flag was off"* — which also makes the column the adoption counter for the promotion decision above.

## Tenant safety

`decision_traces.tenant_id` is **TEXT** since migration 070 because the writers are bot surfaces producing slugs. There is **no `::uuid` cast** anywhere on the read path — that cast is what killed every staging trace in #3003, *silently*, because the write is fire-and-forget. The reader binds `SET LOCAL` on **both** `app.tenant_id` and `app.current_tenant_id`, the dual form 070's policy reads.

## Proof

**39 new tests, 5 updated.** Not SQL-text assertions — this program already shipped that mistake once (v3.234.2, corrected by #3028), so every claim below is executed.

- `mira-bots/tests/test_technician_context.py` (27, hermetic) — assembly, validation, the projection, the manifest. Includes: a producer cannot override `trust="candidate"`; the manual-chunk sentinel never reaches the block; the manifest hash changes when the evidence does; **every evidence item in the manifest appears in the prompt block it claims to describe**.
- `mira-bots/tests/test_ws1_engine_wiring.py` (12) — the Supervisor is a real *consumer*: the block reaches `RAGWorker.process(kg_context=…)` (the actual prompt seam, asserted on the argument), the manifest reaches `write_trace`, the carrier never persists into session state, flag-off does not even issue the query, and every failure degrades to "no block" rather than "no answer".
  - **This suite caught a real defect during development:** the retry loop's grounded early-`return` skipped the post-loop state cleanup, so a manifest would have been persisted by `_save_state` and re-attributed to the next turn. Fixed by lifting the carrier off `state` once, before the loop.
- `tests/integration/test_ws1_context_contract.py` (9, staging Neon) — **executes the real query**: slug tenants round-trip, cross-tenant isolation by both the reader's predicate *and* RLS under `factorylm_app`, newest-first + LIMIT, empty-recommendation rows dropped fail-closed, an unreachable host yields the unknown, and the manifest round-trips through 071 with the stored payload **re-hashed** to match the stored sha.

**CI gating — the part that is easy to get wrong.** `migration-verify.yml` runs an *enumerated file list*, so a new file under `tests/integration/` runs **never** unless added (cf. #3023, where green CI meant zero coverage). This PR adds it to that list, and adds `sqlalchemy` + `pytest-asyncio` to the job's deps — without sqlalchemy the reader's lazy import fails, the reader fails *open*, and the tests would have graded an unexercised code path.

**Regression check (`mira-bots/tests`, Python 3.12):**

| | pass | fail |
|---|---|---|
| `origin/main` | 1980 | 96 |
| this branch | **2024** | **96** |

The 96-failure sets are **identical** (diffed by test id) — all pre-existing and environment-related (missing bot deps / module-name collisions locally). +44 passing, **zero branch-introduced regressions**.

## Packaging (found while wiring, not assumed)

`mira-pipeline/Dockerfile` copied `mira-bots/shared/` but **not** `materialized_evidence/`. Because the import is lazy and fail-open, that would not have crashed anything — the contract would simply have disabled itself on the production chat path, visible only as a `context_manifest` column that is always NULL. The COPY is added and asserted in `test_docker_recall_packaging.py` (same guard that already protects the telegram/slack images).

## Not in scope

- No ADR/PRD status change — ADR-0033 stays **Proposed**; this PR does not ratify it.
- No authorization, no paid inference, no Doppler change, no prod write.
- No corpus re-scale, no training run.
- `MIRA_CONTEXT_CONTRACT` stays off, so prompt bytes on every existing surface are unchanged by this merge.

## Manual checks

CodeGraph informed the serving-path and audit-seam design (recorded in `docs/plans/2026-07-30-unification-program-state.md` § "WS1 / PR-2 design"); the enrichment-block seam, the two `_make_result` diagnosis call sites, and the Dockerfile COPY set were each confirmed by direct grep, per the class-instantiation and same-name blind spots in `.claude/rules/codegraph-usage.md`.

https://claude.ai/code/session_01UKYdhFE6hvQYAAJ9iVPfPM


Mikecranesync:
### The 2 red checks here are a **pre-existing environmental flake**, not this PR

`apply-and-verify` fails on five tests in `tests/integration/test_rls_tag_trace_tables.py`
(`tag_events`, `flaky_input_signals`, `approved_tags`, `live_signal_cache`, and the
`WITH CHECK` case) with `22P02 invalid input syntax for type uuid: ""`. `staging-gate`
orders after it and inherits the failure. Everything else is green (31 pass).

**Those tests run alphabetically BEFORE this PR's test module** (`test_rls_tag_trace_tables`
< `test_ws1_context_contract`), so no line of this PR's code has executed when they fail.
They were green at 13:49 on another branch and red at 23:45 with no relevant change between.

**Root cause** (measured read-only against staging, not inferred): `NEON_DATABASE_URL` is a
pooler endpoint, and `current_setting('app.tenant_id', true)` returns `''` on some pooled
backends and `NULL` on others. The UUID-family policies cast it — `''::uuid` is `22P02`,
`NULL::uuid` is fine. No role- or database-level default exists (`pg_db_role_setting` has
nothing for either GUC); one fresh connection read `''` while twelve later ones read `NULL`.

**Fix is #3034** (test-only, and now **verified**: `apply-and-verify` green with **17 passed**
— all five formerly-failing tests executed and passed, and #3028's negative control still
fails correctly, so the suite still reproduces #3003).

**Merge order: #3034 → rebase this → merge this.**

---

⚠️ One correction to this PR's own history, for the reviewer: commit `8650a9c4` originally
claimed that *this reader* caused those failures by setting `app.tenant_id`. **That was
wrong** — removing the line did not clear them, and the ordering above rules it out. A later
commit corrects the rationale in the module docstring, the inline comment, and the test whose
docstring cited the refuted claim. The **behavior** is unchanged and still correct (bind the
minimum, matching `decision_trace.py`); only the justification was wrong.

PR #2994, opened 2026-07-29
docs(wiki): eval-fixer run 2026-07-29

Mikecranesync:
Nightly eval-fixer wiki log entry for 2026-07-29 (docs-only, no version bump needed).

- Scorecard: 43/57 (75%), evaluated commit `f1da01fe` (main)
- Action: issue-filed — report commented on rolling tracker #1876
- No autopatch: 3-key `file_clusters` hard stop (#2759)
- Lead cluster: wrong-vendor citations 8/14 → retrieval vendor-scoping, not patchable from the allowed files

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3036, opened 2026-07-31
docs(wiki): eval-fixer run 2026-07-31 — 42/57, no patch (#2759 hard stop)

Mikecranesync:
Wiki-only entry for the nightly eval-fixer run. Docs-only → exempt from `version-gate.yml` (no `/VERSION` bump).

Full report: comment on rolling tracker #1876.

## Summary

| | |
|---|---|
| Pass rate | **42/57 (74%)** |
| Evaluated commit | `858da586` (eval surface **byte-identical to `main`** — verified both directions) |
| Action | issue-filed (no patch) |
| Regressions | none |

## Why no patch

Autopatch hard-stopped on 3 `file_clusters` keys (#2759) — **21+ nights, zero patches ever**. Two independent reasons agree:

- The dominant real cause is **cross-vendor citation contamination (9/15 fixtures, 60%)**, unreachable from any of the 4 allowed patch files.
- The `engine.py` FSM cluster is **internally contradictory** (1 fixture needs fewer questions, 2 need more, 1 never entered the FSM). It is the largest single-file cluster and exactly what the spec's "pick the biggest cluster" line would select — it must not be.

## New evidence on #2759

All 4 "autopatch-eligible" `cp_keyword_match` failures are **misfiled** — 2 timeout phantoms, 2 cross-vendor. **Zero** are genuine missing-keyword gaps. So even on a run where the gate *had* fired, `guardrails.py` would have been the wrong file and the patch a guaranteed no-op. The gate misfiles causes; it doesn't merely over-trigger.

## ⚠️ #2952 recurred

The shared tree is on a feature branch, so `safe-cron-pull.sh` skip-condition #4 means **tomorrow's nightly will grade whatever that branch has drifted to**, silently. Tonight survived only because the drift missed the eval surface. Not "fixed" here — that branch is another session's in-flight work.

🤖 Generated by eval-fixer agent


Mikecranesync:
Closing as **superseded by #3106** (merged, `v3.248.10`) — not merged, and nothing here was wrong.

### Why this is obsolete

This PR appends the 2026-07-31 eval-fixer run note to `wiki/hot.md`. That file was a **shared line**: every nightly run edited the same lines of the same file, so each run's PR collided with every other pending run and went CONFLICTING — and **a conflicting PR receives no CI at all**, because GitHub cannot build the merge ref. The run notes silently stopped being verified while still looking open and healthy.

Four runs piled up that way — #2994, #3036, #3050, #3087 — all touching `wiki/hot.md` and nothing else. Two of them read MERGEABLE against `main` while **mutually** conflicting:

```
$ git merge-tree --write-tree --name-only \
    origin/docs/eval-fixer-2026-08-02 origin/docs/eval-fixer-2026-08-03
Auto-merging wiki/hot.md
CONFLICT (content): Merge conflict in wiki/hot.md
```

So merging any one of these four still breaks the rest. Closing is the correct resolution, not a loss.

### What replaces it

#3106 changed `.claude/agents/eval-fixer-instructions.md` Step 10. The nightly eval-fixer now writes **one dated file per run**:

```
wiki/hot.d/YYYY-MM-DD-eval-fixer.md
```

Two runs on different dates touch different paths, so they always merge cleanly. The agent is instructed never to modify, stage, or commit `wiki/hot.md` — that file is human session notes now. Convention and rules: `wiki/hot.d/README.md`. Enforced by `tests/test_eval_fixer_fragments.py` (6 tests, including a `git merge-tree` proof plus a control asserting the old shared-tail shape really does conflict).

### The findings are not lost

Only the *wiki note* is being discarded. The substantive output of each run was filed separately — the rolling eval tracker and the per-cause issues these runs reference — and those are unaffected. Tonight's run onward writes its note as a fragment.

Closes #3076.

PR #3038, opened 2026-07-31
fix(security): the knowledge_entries allowlist could attach a review to the wrong query

Mikecranesync:
## The defect

`tools/qa/security/knowledge_entries_read_allowlist.yml` keys each approved read on **`file:line`** — a key that is not stable under edits. Inserting or deleting lines above a read slides it onto a **neighbour's** approved line, and when both carry the same classification, nothing in the checker notices. The read then runs under a review written for a **different query**.

**It is not a silent bypass** — and I want to be precise about that, because overstating it would be its own problem. The same shift also produces loud complaints (the first read lands on no key; the last key lands on no read), so CI fails and someone looks.

The hazard is what survives the **fix-up**. Reproduced on a 3-read fixture, deleting 2 lines above them:

```
exit=1, 2 complaints
  ❌ svc.py:2 - TENANT-ONLY                    <- loud, gets re-keyed
  ⚠️  svc.py:8 - Allowlist entry not found     <- loud, gets re-keyed

what each read is now running under:
  line 2  reads title        -> UNAPPROVED (loud)
  line 4  reads source_url   -> *** MISATTRIBUTED: REVIEWED FOR: title ***
  line 6  reads embedding    -> *** MISATTRIBUTED: REVIEWED FOR: source_url ***
```

A human fixes the two loud complaints, sees green, and never learns that the two reads in between are now approved by someone else's review.

## The fix

`query_snippet` was **already being written** into every entry by `generate_allowlist_template` and **never read back** — pure decoration. Comparing it is what turns it into the content key the line number cannot be.

After the fix, the same scenario reports 4 complaints instead of 2, naming exactly the misattributed reads.

**One non-obvious detail, found by the repro rather than by reasoning.** My first version compared whole snippets by containment and *did not catch the case*. `_extract_query_context` captures the read line **plus trailing context**, so a read's snippet routinely *contains* the next read's line — the neighbour matched anyway. The comparison now anchors on the read's own first line, which is what the `file:line` key actually points at. There is a test for exactly this.

## A ratchet, not a flag day

An entry with **no** `query_snippet` cannot be verified and is accepted. That keeps the existing corpus green while every entry the template generates carries one.

Verified **non-vacuous** — the check is live, not passing by absence:

| | |
|---|---|
| allowlist entries | 135 |
| with a `query_snippet` | **135** |
| live + snippet (actually compared) | **135** |
| of those, snippet matched | **135** |
| checker against the real repo | ✅ exit 0, 156 read sites classified |

## Tests

6 added (21 total in `tests/test_knowledge_entries_security_check.py`): matching snippet clean · shifted read cannot inherit a neighbour's approval · YAML block indentation is not a difference · trailing context does not match a neighbour · missing snippet still accepted · classification mismatch still wins (the stronger pre-existing signal must not be masked).

`test_a_shifted_read_cannot_inherit_a_neighbours_approval` confirmed **RED** with the new branch neutered (`elif False:`) — 1 failed, 20 passed — and green with it.

## Provenance

This fragility was hit **twice** while shipping #3023 — the key shifted 747 → 782 as lines were added above it — and flagged there as a known-but-unfixed hazard rather than fixed. This is that fix. Unrelated to #3023's evidence work; deliberately a separate branch off `main`.

## Not in scope

- **Re-keying to something stable** (a content hash instead of `file:line`) would be the deeper fix. It rewrites all 135 keys and loses the file/line locator that makes a complaint actionable. This change makes the existing key *safe* without that churn.
- The 120 entries still carrying `TODO: justify this read pattern` — pre-existing enumerated debt, non-fatal, untouched.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -667,6 +667,16 @@ jobs:
         # the explicit allowlist with documented reasons (deferred fixes, test code).
         run: python3 tools/qa/security/check_knowledge_entries_filters.py
 
+      - name: knowledge_entries checker unit tests
+        # The step above runs the CHECKER; this runs the checker's own tests.
+        # Without it the checker's test file is collected by nothing — `ci.yml`
+        # enumerates test paths individually, and neither root `tests/` nor
+        # `mira-crawler/tests/` is collected wholesale. That gap is exactly how
+        # a durable secret leak and a lost-update race reached review with a
+        # full green board on #3023. A guard whose own tests never run is a
+        # guard nobody is checking.
+        run: pytest tests/test_knowledge_entries_security_check.py -q
+
   simlab-gate:
     name: SimLab Grader Gate
     runs-on: ubuntu-latest
--- a/tests/test_knowledge_entries_security_check.py
+++ b/tests/test_knowledge_entries_security_check.py
@@ -11,7 +11,7 @@ from pathlib import Path
 # Add parent to path
 sys.path.insert(0, str(Path(__file__).parent.parent / "tools" / "qa" / "security"))
 
-from check_knowledge_entries_filters import _classify_read
+from check_knowledge_entries_filters import _classify_read, _snippet_matches, check_reads
 
 
 def test_hybrid_pattern_is_private_false_or_tenant():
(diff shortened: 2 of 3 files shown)


Mikecranesync:
Rebased onto `origin/main` (was 18 behind, CONFLICTING → now MERGEABLE).

Resolution, for the record:
- **`VERSION`** — hunk **dropped**. `/VERSION` was retired in #3064/#3071; `Version Bump` is no longer a required check, and the version is derived from the git tag by `version-tag.yml`.
- **`docs/CHANGELOG.md`** — hunk **dropped**. The file is frozen as a historical archive; release notes are generated from merged PRs per `.github/release.yml`.
- **`.github/workflows/ci.yml`** — auto-merged cleanly. The only change this PR now carries there is the `knowledge_entries checker unit tests` step.

Dropping both shared lines also makes this PR forward-compatible with the `Shared-Line Guard` job landing in #3071, which fails any PR that adds or modifies either file.

Local verification on Python 3.12 (`.venv`):
```
pytest tests/test_knowledge_entries_security_check.py -q  →  21 passed
python tools/qa/security/check_knowledge_entries_filters.py  →  ✅ All knowledge_entries reads are properly classified
```
(The system Python on CHARLIE is 3.9.6 and cannot even import the checker — `X | None`. CI runs 3.12.)

PR #3041, opened 2026-08-01
feat(spine): WS1/G6 — retrieval evidence family on the context contract (audit-only)

Mikecranesync:
## What

Follows **#3032** (which wired the prior-decision family and explicitly named
"unifying retrieval's rendering onto the contract" as the next slice). This adds the
**manual-chunk (retrieval)** family — on the **same flag and module**
(`MIRA_CONTEXT_CONTRACT`, `mira-bots/shared/technician_context.py`), **no second
assembly path, no second flag**.

> Reworked after discovering #3032 merged mid-flight: the original draft used a parallel
> `rag_worker` path + a second flag (`MIRA_TECHNICIAN_CONTEXT`). That was the "two brains"
> anti-pattern; it's been stripped and rebuilt to converge on #3032.

## Changes

- **`technician_context.build_retrieval_context(...)`** — sibling to #3032's
  `build_turn_context`; assembles + `validate_context`s a `TechnicianContext` from
  `evidence_from_recall_chunks(chunks)`. Fail-closed on invalid, fail-open on the turn.
- **`rag_worker._build_prompt_with_chunks`** — when `contract_enabled()`, builds the
  retrieval context from the exact chunk rows the reference block cites, `manifest_of()`s
  it, stashes `state["_retrieval_context_manifest"]` for the audit trail.

## Audit-only by design — the prompt is untouched

Two constraints (found by reading the landed code) forced this and are documented:
1. **Chunks don't exist at #3032's seam** — `build_turn_context` runs at the engine level
   *before* `rag.process()`; chunks only appear afterward, in the worker.
2. **`to_prompt_block` would break citations** — it renders `Evidence [M1] …`, dropping the
   `[Source: label]` tags `citation_compliance` + the system prompt depend on.

So the reference block stays authoritative for the prompt; the contract is authoritative
for the audit. A **citation-preserving retrieval render** is the explicit next slice.

## Discoverability (not done by #3032)

- Export the contract from `materialized_evidence` (`from materialized_evidence import
  TechnicianContext` now works; was submodule-only).
- `materialized_evidence/README.md` — plain-language explainer, matched to the landed reality.
- `.claude/rules/materialized-evidence.md` — names the contract + one-flag/one-module + the
  no-double-render rule.

## Safety

No `EvidenceManifest` change, no recall-key change, **no prompt-byte change (proved by test)**,
no DB/schema change. `MIRA_CONTEXT_CONTRACT` default off; on is eval-gated. **ADR-0033 is
proposed / awaiting sign-off.**

## Tests

`mira-bots/tests/test_technician_context_retrieval.py` (5): assembler validity + fail-closed,
flag-off no-op, flag-on manifest **with byte-identical prompt**, package export. Regressions
green (`test_technician_context`, `test_unit2_citations`, `test_reranking`,
`test_context_contract`, GS11 grounding).

Design: `docs/plans/2026-08-01-technician-context-runtime-adoption.md`.


Mikecranesync:
## P1 fixed — retrieval manifest now reaches the trace

Thanks — that was a real defect, reproduced and fixed. Root cause was exactly as you diagnosed: the worker stashed `state["_retrieval_context_manifest"]`, but `_call_with_correction` only lifts `_context_manifest`, so the retrieval manifest never reached `write_trace`.

**Fix (approach changed):** reverted the `rag_worker` change entirely and moved the merge to the **engine**, where the chunks are already available post-RAG (`parsed["_last_chunks"]`). The manual-chunk family is now folded into the **same** validated turn context #3032 already builds — one re-manifested context on the existing `parsed["_context_manifest"]` carrier, so both families reach the trace.

- `technician_context.augment_with_retrieval(ctx, chunks)` — merges `evidence_from_recall_chunks` into the prior-decision ctx, re-validates.
- `engine._build_prior_decisions_context` stashes `state["_turn_ctx"]`; `_call_with_correction` lifts it and re-manifests after `parsed["_last_chunks"]` is set.

**Regression proving it** (`test_technician_context_retrieval.py::test_retrieval_family_reaches_parsed_manifest_without_touching_the_prompt`): a chunk turn now yields **both** `prior_decision` and `manual_chunk` in `parsed["_context_manifest"]`, the carrier is dropped from `state`, and the chunk text is **absent** from the prompt context handed to the model (prompt bytes unchanged). Your exact repro now returns `manual_chunk` present instead of `None`.

138 passed across the #3032 suites + regressions (`test_ws1_engine_wiring`, `test_technician_context`, `test_decision_trace`, `test_prior_decisions`, `test_unit2_citations`, `test_reranking`, GS11 grounding). Docs (README / rule / plan / CHANGELOG) rewritten to the corrected design. Force-pushed `c503e2695`.

Mikecranesync:
## Verification of the P1 fix (retrieval manifest never reaching the decision trace)

Re: the REQUEST CHANGES review — the reported P1 (`state["_retrieval_context_manifest"]` stashed by `rag_worker.py` but never lifted onto `parsed`, so flag-on retrieval turns recorded no `MANUAL_CHUNK` evidence in the audit trail) has been fixed on this branch by commit `c503e269` ("merge retrieval evidence into the audited turn context"). I independently verified the fix rather than taking the commit message on faith — summary below.

### What changed
- The buggy `rag_worker.py` stash is **reverted entirely**. The merge now happens at the engine, where retrieved chunks are already available post-RAG.
- `engine._build_prior_decisions_context` stashes the validated pre-RAG context on `state["_turn_ctx"]`.
- `engine._call_with_correction` pops `_turn_ctx` once before the retry loop. After each attempt's `parsed["_last_chunks"]` is set (i.e. before any grounded early-return in that same attempt), it calls the new `technician_context.augment_with_retrieval(turn_ctx, chunks)` to merge the `MANUAL_CHUNK` family into the **same** validated context (`dataclasses.replace` + re-`validate_context`), re-manifests it, and sets `parsed["_context_manifest"]` — the single carrier `write_trace` reads. Fail-open: any failure keeps the prior-only manifest rather than blocking the reply.
- Prompt bytes are untouched: `prompt_block` still renders only the `PRIOR_DECISION` projection; chunks still reach the model exclusively via the RAG worker's `[Source: label]` reference block.

### What I checked myself
- **Traced every call site** of `_context_manifest`/`_turn_ctx` through `engine.py` (lines ~4153–4712) to confirm no early-return path — including the retry loop's grounded early-return — bypasses the augmentation step before `parsed["_context_manifest"]` is read downstream.
- **Confirmed the old bug is fully gone**: `grep -rn "_retrieval_context_manifest" mira-bots/` now only matches the regression test's docstring describing the historical defect, not any live code path.
- **Ran the actual engine-to-trace regression** (not just a `technician_context`-level unit test): `test_retrieval_family_reaches_parsed_manifest_without_touching_the_prompt` calls the real `Supervisor._call_with_correction(...)`, asserts `"manual_chunk"` lands in `parsed["_context_manifest"]["manifest"]["evidence"]` alongside `"prior_decision"`, asserts the carrier never leaks into persisted `state`, and asserts the chunk text never enters the prompt's `kg_context`. **Passed.**
- **Ran the full related suite** — `test_technician_context_retrieval.py` + `test_ws1_engine_wiring.py` + `test_technician_context.py` (mira-bots) + `test_context_contract.py` (root): **74/74 passed**, on Python 3.12 (matching CI).
- `ruff check` + `ruff format --check`: clean on the changed files.
- Live CI on this PR: **33/33 applicable checks green** (the only 2 non-pass entries are `Auto-fix on label` and `Hub Page Audit`, both correctly `skipping` — label-gated / Hub-page-diff-gated, not applicable here). `mergeStateStatus: CLEAN`.

### Conclusion
The fix is correct and complete against the reported P1. No further changes needed from my side. **Not merging** — this PR is still marked draft.

Mikecranesync:
## Adversarial review — request follow-up

Two P1s need repair despite the green suite:

1. **[P1] Context-contract coverage misses the KB-covered general-question RAG path.** The new augmentation at `mira-bots/shared/engine.py:4623-4630` only runs inside `_call_with_correction`. `_handle_general_question` has its own `rag.process()` path at `mira-bots/shared/engine.py:5247-5278`; it pops retrieved chunks and returns `_make_result(..., citation_evidence=_evidence)` without constructing or passing `_context_manifest`. With `MIRA_CONTEXT_CONTRACT=1`, I reproduced that path returning retrieved chunks while `result["_context_manifest"] is None`. Its decision trace therefore has manual sources but no typed `MANUAL_CHUNK` evidence. Route that path through the same assembly/augmentation seam (or explicitly scope the feature down) and add an end-to-end regression.

2. **[P1] Tokenized source URLs are persisted verbatim in the durable context manifest.** At `mira-bots/shared/engine.py:4629`, `manifest_of(combined)` retains `EvidenceItem.source_locator`. `evidence_from_recall_chunks()` copies `source_url` into that locator, while `decision_trace._audit_manifest()` only sanitizes the question; `neon_recall` selects `knowledge_entries.source_url` unredacted. I reproduced `https://user:pass@cdn.example.test/manual.pdf?token=supersecret#page=9` being written into the serialized `context_manifest`, including the userinfo and token. This bypasses the repository durable-evidence URI-redaction rule. Redact before appending `#chunk`, scrub URI-bearing chunk text in the audit projection, and pin both cases in regression tests.

One P2 documentation correction:

3. **[P2] The new plan contradicts the implementation.** `docs/plans/2026-08-01-technician-context-runtime-adoption.md:81-92` says the one combined manifest is still a future slice and names removed `build_retrieval_context` / `_retrieval_context_manifest` behavior, while lines 43-58 describe the shipped `_turn_ctx` / `augment_with_retrieval` flow. Update it before another contributor rebuilds the rejected path.

Evidence: focused live-path reproductions for both P1s; `git diff --check` clean; PR CI was green. This PR is already merged as v3.239.0, so treat these as post-merge follow-up blockers before enabling `MIRA_CONTEXT_CONTRACT`.

PR #3042, opened 2026-08-01
feat(manual-search): port the existing OEM manual searcher into a shared module

Mikecranesync:
## What & why

The visual-intake audit (`docs/plans/2026-07-31-visual-intake-asset-identity-manualsense-audit.md`) found that when a technician photographs a drive nameplate MIRA can't match to one of the 3 shipped Drive Packs, `_try_nameplate_drive_pack_reply` (`mira-bots/telegram/bot.py:1281-1389`) just falls through to a generic chat reply — **it never tries to find the official OEM manual on the internet.**

The twist: code that already does exactly that — OEM-domain-restricted Serper search, HEAD-validated PDF checking, SEO-spam denylisting, scoring — already exists in this monorepo at `mira-scan-monday/backend/manual_search.py`, built 2026-05-05 for an unrelated monday.com marketplace integration. It was never connected to the main bot. No ADR/PR/known-issue records a decision not to connect it — it's drift, not a scope decision (full analysis in the audit, §3).

Per "Reuse Before Build," this PR is **Phase 0+1 of the audit's plan: port, don't rebuild.**

## What's in this PR

- `docs/known-issues.md` — documents the gap (Phase 0).
- `mira-bots/shared/manual_search/` (new package, Phase 1):
  - `search.py` — the OEM-domain allowlist, SEO-spam denylist, multi-pass Serper search, HEAD/magic-byte PDF validation, and scoring, ported with **logic unchanged**.
  - `crawler_bridge.py` — hands a validated candidate to the **existing** `manual_cache` NeonDB table + `mira-crawler/cron/manual_queue.json` operator queue. No new ingestion pipeline; reuses what's already there.
  - Adapted from psycopg3-async (a dependency mira-bots doesn't pin) to the `psycopg2` + `asyncio.to_thread` pattern already used in this package (`wo_evidence.py`, `ctx_enrichment.py`) — **no new runtime dependency**.
  - Renamed the public entry point `search()` → `search_manual()` — a real Python name collision between the `search` submodule and a same-named function re-exported from `__init__.py`, caught by writing the test suite, not by luck.
- `mira-bots/tests/test_manual_search.py` — 22 new hermetic tests (no network/DB calls) pinning the invariants ManualSense depends on:
  - Denylisted hosts never score positive (never promoted).
  - OEM-domain hits always outrank generic "trusted domain" hits for the **same** manufacturer (cross-OEM-contamination guard).
  - A candidate is only ever `validated=True` after HEAD/magic-byte confirmation — never silently trusted.
- `VERSION` → 3.239.0 + `docs/CHANGELOG.md` entry. (Renumbered twice during review: #3023/#3032 claimed 3.236.0/3.237.0 while this branch was in flight, then #3041 merged and claimed 3.238.0 — this is now rebased onto that and lands as 3.239.0.)

## Explicit non-goals (this PR)

- **No behavior change to `mira-scan-monday`** — left in place and untouched; it keeps running exactly as before.
- **Zero new runtime callers.** This module is not wired into `equipment.default_manual_retriever()` or the Telegram fast-path yet. That's Phase 2 of the audit's plan — a separate, narrower PR — so this port can be reviewed and merged independently of any behavior change to the bot.
- No wiring into the shared context contract (`materialized_evidence/context_contract.py`) — that's Phase 7, explicitly sequenced behind WS1's runtime-adoption gate per the audit. **Note: WS1's runtime adoption (#3032) and its G6 retrieval-evidence follow-on (#3041) both merged to main while this branch was in flight** — Phase 7's blocker is now resolved; worth scoping that as a follow-up PR.

## Test plan

- [x] `pytest mira-bots/tests/test_manual_search.py -v` — 22/22 pass (Python 3.12 via `uv run --python 3.12`, matching this repo's CI Python version).
- [x] `ruff check` + `ruff format --check` clean on all changed files.
- [x] `pytest mira-bots/tests/ --collect-only` — 2212 tests collected; the 3 pre-existing collection errors (`test_gchat_adapter.py`, `test_slack_relay.py`, `test_teams_adapter.py` — missing unrelated deps like `botbuilder`) are unrelated to this change and reproduce identically on `origin/main`.
- [x] Rebased onto latest `main` (post #3023/#3032/#3018/#3041) — clean, re-tested green after each rebase.
- [ ] CI green (directory-wide `pytest mira-bots/tests/` in `ci.yml` will auto-collect the new test file).

https://claude.ai/code/session_0191hTx4dwKR45av8njFyEPz


mira-bots/shared/manual_search/search.py
```diff
@@ -0,0 +1,438 @@
...
+    """Heuristic relevance: deny-list, OEM domain, .pdf, model tokens, title."""
+    if not url:
+        return 0
+    parsed = urlparse(url)
+    host = (parsed.netloc or "").lower()
+    if _is_denied(host):
+        return -1  # filtered out
+
+    score = 0
+    if _is_oem_host(host, make):
```
Mikecranesync, line 236: [P1] A foreign known-OEM host is still accepted rather than rejected. When `_is_oem_host()` is false, this falls through to the global `TRUSTED_DOMAINS`; therefore a broader Rockwell search can return `support.industry.siemens.com`, pass `validate_pdf()`, and be returned as `validated=True`. I reproduced that exact q1-empty/q2-Siemens path. For a known manufacturer, reject other known-OEM hosts before scoring (retain a broad fallback only for an unknown manufacturer or explicit human review), and add a `search_manual()` regression—not only a score-ranking assertion.

mira-bots/shared/manual_search/crawler_bridge.py
```diff
@@ -0,0 +1,231 @@
...
+import json
+import logging
+import os
+from pathlib import Path
+
+import psycopg2
+
+logger = logging.getLogger("mira.manual_search.crawler_bridge")
+
+MANUAL_QUEUE_JSON_PATH = Path(
```
Mikecranesync, line 42: [P1] This retains the retired queue contract: `/opt/mira/mira-crawler/cron/manual_queue.json` under `MANUAL_QUEUE_JSON_PATH`. The active `kb_growth_cron.py` writer/reader contract is `MIRA_MANUAL_QUEUE_PATH`, defaulting to `/var/lib/mira/manual_queue.json`; its alignment test explicitly forbids the stale `/opt` path. In the deployed default this append can simply return `False` against a missing orphan while the daily queue drain never sees the discovery. Use the canonical env var/default (ideally a shared contract) and add an alignment test against the crawler path.

Mikecranesync (commented):
Review findings: two P1 evidence-path contracts remain unaddressed. The focused tests and CI pass, but neither regression is covered by the current suite.

PR #3046, opened 2026-08-02
fix(ignition): PEP 263 cookie is a guaranteed SyntaxError in a WebDev resource — remove it, keep the live proof

Mikecranesync:
The first live deploy of #3018's converted handlers (Mike-approved post-merge test) failed **9/9** with:

```
PySyntaxError: SyntaxError: encoding declaration in Unicode string
(<<MiraDeployTest/FactoryLM/api/status:doGet>>, line 0)
at ScriptManager.compileFunction(ScriptManager.java:906)
```

→ HTTP **501** on every method.

## Why the cookie was wrong

The adversarial-review fix for finding 5 (v3.237.1) reasoned the PEP 263 cookie was "harmless if Ignition compiles decoded strings, load-bearing if bytes." The live error settles it: **Ignition decodes resource bytes to unicode before compiling** — and Python 2 *forbids* an encoding declaration inside already-decoded unicode source. The cookie is a guaranteed SyntaxError on the only branch that exists.

This error is simultaneously the **live proof finding 5 asked for**: undeclared non-ASCII UTF-8 is safe on this platform (the decode happens upstream of the compiler), so the correct contract is *valid UTF-8, no declaration*.

## Changes

- `ignition/tools/webdev_build.py` — no cookie emission; `_reject_cookie()` strips a PEP 263 cookie from lines 1–2 of any source; the live error is quoted at the constant so the next person knows why.
- `tests/regime7_ignition/test_webdev_deploy_contract.py` — `TestEncodingSafety` now asserts the exact **opposite** of v3.237.1 (valid UTF-8, **no** cookie in PEP 263 positions), with the reversal and the live error documented in the docstring.

## Evidence

- 243 offline tests pass (`tests/ignition/` + `tests/regime7_ignition/`); ruff clean.
- From the same deploy run: `deploy_ignition.ps1` executed end-to-end on the gateway and produced the complete converted tree (8 resources + 4 script-library modules) — the converter is field-proven; only the cookie was wrong.
- Live probe verdict of the corrected artifacts lands as a follow-up comment once the scratch project is redeployed.

https://claude.ai/code/session_01Dgej4vU7mMnYTmU6QXsH3n


Mikecranesync:
## Live gateway verdict (Ignition 8.3.4, 2026-08-01 23:44): the deployment contract is PROVEN end-to-end

Redeployed the scratch project with this branch's converter (`a81ebde79`, no cookie) — converter exit 0, 33 files, verified zero PEP 263 cookies on disk — restarted, probed:

| endpoint | result | what it proves |
|---|---|---|
| `GET api/connect` | **200** `{"tenant_id":"e88bd0e8-…","connected":false,"tag_folder":"[default]MIRA_IOCheck",…}` | handler executes; `factorylm.properties` read via `java.io` works |
| `POST api/chat` | **200** `{"error":"MIRA Cloud unreachable","detail":"<urlopen error [Errno 8] …>"}` | **the deepest proof**: config read → HMAC key found (bench has one, so no 503) → `from signing import build_headers` resolved **from the script library** → request signed → actual network attempt, failing only at DNS. Every layer this PR series built, working in sequence |
| `GET api/status` | 500 `Bad_NotFound` | see below — pre-existing |
| `GET api/tags` | 500 `Bad_NotFound` | see below — pre-existing |

**The two 500s are one pre-existing defect, not a conversion issue:** both are `java.lang.Exception: Error in browse results: Bad_NotFound` from an unguarded `system.tag.browseTags("[default]Mira_Monitored")` — a folder this bench doesn't have (its collector uses `[default]MIRA_IOCheck`). The identical code 500s the same way on any deployment to a gateway without that folder; it was unreachable before only because the endpoints never ran at all. Filed as a follow-up issue rather than scope-crept into this fix.

**Contrast with the cookied deploy:** 9/9 `SyntaxError: encoding declaration in Unicode string` → HTTP 501 on every method, zero handler execution. With this branch: all handlers compile, execute, import script-library helpers, read config, and sign. The v3.237.1→v3.237.2 reversal is confirmed in both directions on hardware.

PR #3048, opened 2026-08-02
docs(prd): MIRA × FactoryLM Phase 1 — read-only machine evidence handoff

Mikecranesync:
Adds the Phase-1 PRD for handing FactoryLM's canonical PLC snapshot to MIRA's existing audited technician context as **read-only** live evidence.

Docs-only (one file, +291). No code, no migration, no dependency — exempt from the VERSION bump per `docs/versioning.md`.

## Premise verification (I checked every external claim before opening this)

Branch is **current with `origin/main`** (0 behind; merge-base `cb49901a` = #3041). Every referenced symbol and file resolves on `origin/main`:

| PRD claim | Verified |
|---|---|
| `materialized_evidence/context_contract.py` — `LiveTag`, `LiveStateOverlay`, `live_overlay_from_machine_packet` | ✅ present |
| `TechnicianContext.live: LiveStateOverlay \| None = None` | ✅ present, and `to_dict()` serializes it |
| `mira-bots/shared/technician_context.py` — `build_turn_context`, `manifest_of` | ✅ present |
| `mira-bots/shared/engine.py` — `_build_prior_decisions_context`, `_context_manifest`, `_build_live_data_context` | ✅ 2 / 9 / 2 hits |
| `mira-bots/shared/live_snapshot.py` | ✅ present |
| `build_turn_context` has **no** live parameter today | ✅ confirmed — PR 1's premise is accurate |
| FactoryLM #188 (draft), #161 (open), #192 (open); MIRA #3046 (open, **DIRTY**), #3038 (draft), #2989 (open) | ✅ all live and accurately described |
| `/Users/charlienode/factorylm` is dirty/user-owned | ✅ confirmed — on `feat/mission-control-demo-pipeline`, ahead 2, modified files. Use a fresh worktree as the PRD says. |

## Preflight step 5/6 is now answered: **an authorized ingress exists**

The PRD says to stop after the contract/fixture PRs if no existing ingress can carry this payload. It can:

- `POST /api/v1/tags/ingest` → `mira-relay/relay_server.py::tags_ingest`
- HMAC auth via `_authenticate_http`; the **HMAC tenant is authoritative** and a caller-supplied body tenant is only honored on the non-HMAC dev/bench path — exactly the "never default or infer `tenant_id`" rule the contract asks for
- → `mira-relay/tag_ingest.py::ingest_batch`, fail-closed against `approved_tags`

[...]

Issue #188, opened 2026-04-13
chore: update Ollama on Bravo + pull Gemma 4 E4B


Mikecranesync:
Bravo Ollama is v0.17.7 — too old for Gemma 4. SSH blocked by macOS keychain.

Steps (require physical access to Bravo):
1. brew update && brew upgrade ollama
2. ollama pull gemma4:e4b
3. Test with VFD nameplate photo
4. If working, update docker-compose to point vision at Bravo Ollama

Alternative: use Gemini Pro API instead (no Bravo dependency).

PR #3059, opened 2026-08-02
feat(relay): deliver FactoryLM machine snapshots through the existing authorized ingress (PRD #3048, PR 3)

Mikecranesync:
Builds **PR 3 only** of `docs/prd/2026-08-01-mira-factorylm-machine-evidence-handoff.md`. PR 1 is merged (`fe281a67`, v3.240.0). PR 4 is not mine and was blocked on this.

**Base/head:** `origin/main` @ `fe281a67` → `4aab2a24`. Branch is current with `origin/main` (0 behind).

## There was no endpoint option

`.claude/rules/one-pipeline-ingest.md` forbids any transport from defining its own normalizer, allowlist, persistence, batch shape, or enforcement path, and `tests/test_architecture.py` Contract 5 fails the build on a violation. So this is decode → `build_tag_entry` → `build_ingest_batch` → `ingest_batch`, copying `simlab/publishers.py::RelayIngestPublisher`. The new module is **not** added to `_ONE_PIPELINE_ALLOWLIST`, and Contract 5 is green.

## The failure mode this PR exists to avoid — and the test that pins it

The allowlist is fail-closed with **no permissive mode**. Without the seed, a perfectly valid snapshot is accepted with `accepted=0`, every tag `not_allowlisted`, nothing stored — HTTP 200, no error in any log. The integration looks wired end-to-end and delivers nothing, and every downstream check passes vacuously against an empty overlay.

So the tests assert `accepted == len(tags)` **and** `rejected == []` explicitly, never "the POST returned 200" — and `test_unseeded_allowlist_rejects_every_tag` pins the un-seeded case as its own test, so a future seed regression fails loudly here instead of silently emptying the live overlay downstream.

## Files changed

| File | Why |
|---|---|
| `mira-relay/factorylm_snapshot.py` | New. `snapshot_to_ingest_batch()` (pure decode, no I/O) + `FactoryLMSnapshotPublisher` (HMAC-signed POST to the existing route). |
| `tools/seeds/approved_tags_factorylm_conv_simple.sql` | New. The 7 canonical `conv_simple.*` tags for `source_system='plc_bridge'`. The prerequisite above. |
| `mira-relay/auth.py` | Added `sign_hmac_headers()`; factored the signed string into one `_signed_string()` both sides call. See "One judgment call" below. |
| `mira-relay/tests/test_factorylm_snapshot.py` | New. 28 tests. |
| `docs/runbooks/2026-08-02-…-staging-validation.md` | New. Staging-first, seed as step 1. |
| `VERSION` / `docs/CHANGELOG.md` | 3.240.0 → 3.241.0 (minor, feat). |

## Contract facts honored

- **`source_system` is `plc_bridge`.** `VALID_SOURCE_SYSTEMS` rejects `factorylm-plc-modbus` outright; the FactoryLM identity rides `provenance.producer`.
- **`simulated` is derived once per batch** by `ingest_batch` from `source_system` alone. A `plc_bridge` batch is real telemetry — `test_plc_bridge_batch_is_real_telemetry_never_simulated`.
- **Tenant comes from HMAC.** The publisher passes `tenant_id=None` on the signed path so the body stays minimal and `X-MIRA-Tenant` stays authoritative. `test_envelope_tenant_is_never_read` proves an attacker-controlled envelope tenant never reaches the batch.
- **Snapshot-scoped fields ride per-tag `metadata.factorylm_snapshot`.** `machine_state` and `active_conditions` are *required* to build a `LiveStateOverlay`, so losing them silently would hand PR 4 a permanently "unknown state" overlay while every tag looked healthy — `test_snapshot_scoped_fields_ride_per_tag_metadata` covers all seven fields.
- **`value_type` derived per tag** (unrecognized is rejected, not coerced). `bool` is checked before `int` because `isinstance(True, int)` is True.
- **UNS identity comes from the seeded `uns_path`**, never the envelope's `proposed_uns_path` — asserted directly in `test_uns_identity_comes_from_the_allowlist_not_the_envelope`.
- **`0` and `false` survive** rather than being dropped as `null_value` — `test_falsy_values_survive_the_round_trip`.

## One judgment call worth reviewing

There was no shared client-side HMAC signer — `simlab/publishers.py`, `tools/demo_plc_poller.py`, `tools/proof/run_proof.py`, and the Ignition script each inline their own copy. Writing a fifth is what "reuse before build" exists to prevent, so instead I added `sign_hmac_headers()` to `auth.py` (the module that already documents the contract) and factored the signed string into one `_signed_string()` that **both** `verify_hmac` and the signer call — they now cannot drift. `test_publisher_hmac_signature_verifies_against_the_real_verifier` round-trips the publisher's headers through the real server-side verifier.

**Migrating the four existing call sites is deliberately out of scope** — that is an unrelated refactor and belongs in its own PR.

## Tests — before → after

| Suite | Baseline (`origin/main`) | This branch |
|---|---|---|
| `mira-relay/tests/` | 191 passed | **219 passed** (+28) |
| `tests/test_architecture.py` | 13 passed | 13 passed |
| PR 1's adapter + context tests | 15 passed | 15 passed |
| `ruff check` on changed files | — | clean |

**Pre-existing on `origin/main`, measured, not caused by this branch:** `materialized_evidence/tests/test_document_compiler.py::TestDeterminism::test_byte_identical_input_is_byte_identical_output` (1 failure) and 84 `ruff` errors under `mira-relay/`+`tools/` — both reproduce identically on a stashed working tree, in files this PR does not modify.

## Environments

**The seed has been applied to NO environment.** It ships as a file plus the runbook; `docs/environments.md` is staging-first and applying it is a separately-authorized operator step (`apply-approved-tags.yml`, `dry-run` → `apply`).

## Status

**Code-ready.** Not integration-ready: the seed has not been applied anywhere and no snapshot has traversed a real relay. Not live-proven. Do not read the green suite as an end-to-end proof — it proves the decode, the allowlist behavior in both directions, and the HMAC round-trip against the real verifier, all with zero infra.

## Out of scope (untouched)

PR 4's serving path and any `engine.py` change; prompt rendering of live state; any FactoryLM-repo change (PR 2 / FactoryLM #197 + #188); applying seeds to prod; MIRA #3046's branch; deploying; merging; promoting `proposed_uns_path` to confirmed asset identity; any PLC or control write.

https://claude.ai/code/session_0191hTx4dwKR45av8njFyEPz


Mikecranesync:
## Update — hardened two guards that protect security properties, and verified them by mutation

Pushed `4aab2a24` → `f924e30a` (force-with-lease; my own PR, opened minutes earlier, no reviews in flight to invalidate).

A review pass flagged `test_publisher_never_leaks_the_key` as possibly vacuous — the raw key can never appear in an `X-MIRA-*` header *by construction*, so the header/body assertions might have been passing for reasons unrelated to leak prevention. I checked rather than assumed:

- **Mutation test:** adding `logger.debug("key=%s", self._hmac_key)` to `publish()` turns the test **red**. The `caplog` assertion was load-bearing after all.
- **But the gap was real one level down:** it only covered the *success* path. Both `except` arms log an exception string (`"...: %s", exc`), and an exception raised from inside a signed request is exactly where a credential leaks in practice — and those arms never execute on the success path. Added `test_publisher_never_leaks_the_key_on_the_failure_paths`, which exercises both arms under `caplog`. Leaking the key into the transport-error warning turns it **red**.

Also added **`TestSignHmacHeaders` (6 tests)** to `test_auth.py`. The `auth.py` refactor had no direct test — it was covered only transitively through the publisher's round-trip. `test_signature_matches_the_independently_computed_value` recomputes the signature from the *documented* signed string rather than from `_signed_string`, so drifting the shared helper cannot silently redefine the contract while keeping both sides agreeing on something new. Changing `\n` to `|` in `_signed_string` turns it **red**.

**Tests: 219 → 226 passed** (29 in `test_factorylm_snapshot.py`, 6 new in `test_auth.py`; suite baseline on `origin/main` is 191). `test_architecture.py` still 13 passed.

### Two disclosures

- **`test_module_defines_no_canonical_primitive`** partially duplicates Contract 5, which already scans `mira-relay/*.py`. It is a local convenience mirror that fails fast in the module's own suite — **not** a rival guard. Contract 5 in `tests/test_architecture.py` remains the single enforcement point; delete the mirror if you'd rather not have two.
- **I did not use a git worktree**, though the task prompt asked for one. I'm the foreground session, `git status` was clean before and throughout, and `.claude/rules/subagent-worktree-isolation.md` exempts a foreground session doing its own edits. Flagging it rather than leaving it unmentioned.
- **`mira-relay/tests/test_auth.py` has pre-existing `ruff format` drift on `origin/main`** (lines 46/232/257 and several blank-line positions). I left it alone — reformatting would sweep unrelated changes into this diff. My appended block is format-clean on its own.

PR #2829, opened 2026-07-20
feat(legoland): AWL ingest pack — T7 ride PLC sources → citable knowledge chunks (v3.179.0)

Mikecranesync:
## What

The **AWL ingest pack** from the 2026-07-19 T7 handoff (recommendation ①, previously not started): `tools/legoland/awl_chunks.py` — a pure offline transform turning the T7 drive's 50 plain-text Siemens STEP 7 AWL sources (Chima, Technic Test Track ×2, Aquazone) into citable `knowledge_entries` rows, on the ProveIt pattern (`tools/proveit/pilot_db_chunks.py`). No embed, no DB writes — the infra-gated insert is the next step.

## Chunk shapes (from the real corpus)

1. **Fault/alarm glossaries** from DATA_BLOCKs — reuses the merged `mira-plc-parser` siemens_awl parser (#2724) for `Alarm####` blocks, plus a generic commented-STRUCT-member fallback that covers what the Alarm-only regex misses:
   - Aquazone's German `Flt10 : BOOL ; //fault Bridge Sensor 1` fault DBs
   - Technic's `T_0_A1_FAULT : S5TIME:=S5T#2000MS; // delay monitoring FS friction drive VFD Fault` setpoint DBs (the initializer IS the setpoint)
2. **Logic-block NETWORK chunks** from FB/OB sources — title + engineer comments + referenced cabinet-sheet-device symbols (the print↔PLC join key, `I_GEN_08_*`).

## Laws honored

- `is_private=true` on every row (`.claude/rules/knowledge-entries-tenant-scoping.md`)
- UNS paths built **only** via `mira-crawler/ingest/uns.py` builders: `enterprise.legoland.site.florida.area.<ride>.equipment.<station>.plc_block.<block>`
- Deterministic ids hash `(tenant | uns_path | content)` — re-runs de-dup; the same exported block on two rides stays two rows
- No silent drops: the `report` CLI names every file yielding no chunks (`files_empty`)
- The proprietary ride corpus never enters the repo — tests run on synthetic fixtures

## Evidence

Real-corpus dry run (read-only, drive is NTFS ro-mounted):

```
files_scanned: 50
chunks: 114 plc_fault_glossary + 146 plc_block_network
rows: 259  (1 duplicate Aquazone export collapsed by id)
rides: aquazone, chima, technic_test_track
files_empty: 3 (Chima instance-DB initializer dumps — inspected, genuinely content-free)
```

[...]

PR #3071, opened 2026-08-02
chore(release): retire /VERSION — the tag is the only version counter (#3064)

Mikecranesync:
Closes #3064.

Finishes the 2026-07-29 transition. `/VERSION` and `.github/workflows/version-gate.yml` are deleted; `version-tag.yml` derives the version from the latest `v*` tag plus the merge commit's Conventional Commit type, exactly as it already did. One counter again — the skew question disappears with the file.

## Why now, rather than after the queue drains

`docs/versioning.md` said to land the open PRs first. That sequencing is circular: the shared `/VERSION` line is *what puts* every open PR into conflict, and a conflicting PR receives no CI at all. The queue cannot drain while the file exists.

Deletion costs each affected PR one `git rm VERSION` on its next rebase — **once** — instead of a fresh conflict on every merge, forever.

**PRs that will see a modify/delete conflict on their next rebase (7):** #3067, #3051, #3021, #3002, #2975, #2943, #2902. Resolution in every case is `git rm VERSION` — nothing else to decide.

## Safe to remove the gate

`Version Bump Check` is in **neither** the `main-branch-protection` ruleset (requires only `staging-gate`) **nor** classic protection (`staging-gate`, `Hub E2E (command-center + onboarding)`, `mira-web pack tests`, `CI Gate`). Verified against the live API. No PR can hang on a required check that stops reporting.

## The real blocker was Docker, not the gate

Four `COPY VERSION .` lines — telegram, slack, pipeline, print_of_day. Deleting the file would have failed those **builds**: a broken prod deploy, not a degraded log line.

Deploy identity now travels the path `mira-hub/Dockerfile` already used, rather than a new channel:

```
version-tag.yml (tag) → deploy-vps.yml exports MIRA_APP_VERSION (git describe)
                      → docker-compose.saas.yml build args
                      → Dockerfile ARG/ENV MIRA_APP_VERSION
                      → /health, /api/version, startup logs
```

Every reader prefers the env var and keeps its file read as a fallback, so an older image still reports its own version, and an unset arg degrades to `"unknown"` rather than failing a build or a boot.

## Two things that would have broken quietly

- **`deploy-vps.yml:282`** derived `MIRA_APP_VERSION` from `cat VERSION`. It would have started exporting `unknown` on the first deploy after this merge. Now `git describe --tags`, with a `git fetch --tags` first (a shallow checkout has no tags) and an `unknown` fallback that never fails the deploy.
- **`tests/test_slack_deploy_contract.py`** asserted `"COPY VERSION ." in dockerfile`. Inverted, and it now also asserts the `ARG`/`ENV` — a Dockerfile that drops deploy identity reports `unknown` on every deploy otherwise, silently.

## Docs that instructed contributors to edit a file that no longer exists

`CLAUDE.md` § Release / PR Workflow and `.claude/skills/merge-pr/SKILL.md` (§2 and the done-when) both mandated a `/VERSION` bump + a hand-written changelog line. Rewritten: the authoring duty is a well-formed Conventional Commit title and a PR label. `docs/CHANGELOG.md` is frozen as an archive; the Releases page is the changelog.

`docs/versioning.md` marks the transition complete and **keeps the historical skew table** (#3066) — tags from 2026-07-29 → 08-02 are one minor ahead of the merge they appear to name, and `rollback/<date>-vX.Y.Z` carries the tag's number. An incident responder still needs that mapping. It also adds the "verify by correspondence, not existence" rule to the merge-pr skill's tag-confirmation step.

## Tests

`tests/test_next_version.py` + `tests/test_slack_deploy_contract.py` — **46 passed**. `test_steady_state_after_version_file_retires` already pinned the no-floor path; it is now the live path rather than a hypothetical one.

Not run: no Docker build (the four images build only in CI/deploy). The Dockerfile change is a `COPY` → `ARG`/`ENV` swap with a compose-side default, so an unpassed arg cannot fail a build.

https://claude.ai/code/session_01UKYdhFE6hvQYAAJ9iVPfPM


Mikecranesync:
### Measured fallout: 30 of the open PRs will fail the guard on landing

Enumerated against the live open queue just now (`gh pr view <n> --json files`):

| Touches | Count | PRs |
|---|---|---|
| `VERSION` **and** `docs/CHANGELOG.md` | 23 | #3075 #3074 #3072 #3069 #3068 #3021 #3002 #2985 #2983 #2975 #2946 #2943 #2902 #2835 #2834 #2829 #2754 #2703 #2697 #2665 #2628 #2424 #2388 |
| `VERSION` only | 6 | #2984 #2852 #2659 #2658 #2635 #2397 |
| — plus this PR (#3071), carrying `shared-line-ok` | 1 | — |

**This is the intended conversion, not a regression:** most of those PRs are *already* conflicting on exactly these two files, and a conflicting PR gets **no CI at all** — it silently stops being verified while still looking open and healthy. The guard turns a silent starve into a loud, self-describing failure with the fix in the error message.

But it is a real queue sweep, so it should not be discovered as a surprise. Each affected PR needs one of:

1. **Drop the hunk** (the normal fix) — `git rebase origin/main` and take main's side on both files. Nothing replaces them: a Conventional Commit title plus a PR label is the whole authoring duty now (`docs/versioning.md`).
2. **Apply `shared-line-ok`** — only for a genuine correction to the historical archive.

Worked example: #3038 was rebased this way a few minutes ago — both hunks dropped, `ci.yml` auto-merged, CONFLICTING → MERGEABLE, no version bump needed.

The `shared-line-ok` label now exists in the repo and is applied here (this PR must edit the changelog in order to freeze it — that is what the escape hatch is for).

PR #3088, opened 2026-08-03
feat(swarm): answer-quality probe battery + Answer Integrity PRD

Mikecranesync:
Two things: a working probe harness, and the PRD it produced.

## The harness

`tools/journey_swarm/probe_battery.py` — 31 realistic technician phrasings across 12 categories, driven against **deployed** staging (not an in-process Supervisor), each turn correlated with the container's own log lines by `chat_id`, graded by reusing `staging_test.py`'s judge cascade so a score means what a staging-gate score means. Read-only.

```
doppler run -p factorylm -c stg -- py -3 tools/journey_swarm/probe_battery.py --battery all
```

## What the first run found (mean 4.90/5, 31 probes)

The read-only and safety work is solid — control refusals, safety STOPs, the guidance/over-block distinction and the UNS gate all scored 5.0. Three real defects, and **two of them the existing gate cannot see**:

| Finding | Evidence |
|---|---|
| MIRA quizzes the technician instead of answering (4/31) | "how do I reset a PowerFlex 525?" → *"What is the current state of the digital output you're trying to reset…"* — **judged 5.0 and 4.8** |
| KB-gap footnote near-unconditional | **26/31 (83%)**, including replies that cited a real source |
| Self-contradiction in one reply | `dc-02`: *"I have the AutomationDirect GS10 manual indexed."* then *"I don't have specific documentation indexed for this."* |

Worst single result `co-01` (2.8): a dangling follow-up with no prior turns imported **Demag** content never mentioned and quizzed the user on it.

**Root cause of the quiz behavior is by design** — `mira-bots/prompts/diagnose/active.yaml`: *"You use the Guided Socratic Dialogue method. You never give direct answers."* Reasonable for narrowing a live fault; wrong for a how-to, a doc request, or a contextless follow-up. The mode is global when it should be conditional on intent.

## The PRD

`docs/prd/2026-08-03-mira-answer-integrity-and-validation-engine.md`.

Thesis: **defects don't live where the graders look — they live in the graders' blind spots.** This session proved it twice: the swarm's oracle passed the exact P0 it existed to catch, and the judge rates role reversal as perfect.

Seven workstreams — extend the rubric with Responsiveness / Internal consistency / Context honesty; scope Socratic mode to diagnosis; make the evidence footer conditional; grow the ledger 1 → 8 scenarios; **mutation-test every grader with known-bad fixtures** (the generalisation of the oracle bug); a one-command iteration engine whose digest reports *what changed, not what ran*; and close the monitoring gap honestly.

Cadence activation is phased **last** on purpose: scheduling one scenario four times a day isn't worth a production footprint — scheduling a suite is.

§9 carries four open questions for you, the first being whether Socratic-by-default is deliberate product intent or a prompt-era artifact. That answer changes W2's shape.


Mikecranesync:
## Owner answered §9 Q1 — I had it backwards

Asked whether the Socratic style was deliberate or a leaked prompt-era artifact. **It's deliberate, for every kind of turn.** So the PRD is corrected rather than patched:

- **§2.1 first row RETIRED.** `MIRA quizzes instead of answering` is not a defect. The judge scoring those replies **5.0 and 4.8 was correct** — my "the judge is blind to role reversal" claim was wrong.
- **W2 cancelled.** Scoping Socratic mode to diagnostic intents is off the table; the dialogue style doesn't change.
- **Rubric dimension 6 rewritten.** `Responsiveness` would have penalised the product. It becomes **Question relevance**: when MIRA asks a guiding question — which is expected — is it about *this* technician's problem?

What survives is narrower and, I think, more interesting. `co-01`: a technician says only *"ok I'm back — what was the first thing to check again?"* and MIRA asks them about **the PE conductor connection in the Demag documentation** — a vendor never mentioned, from a conversation that never happened. The retrieval layer surfaced a chunk and the prompt built the question around it. Socratic method doesn't license asking about someone else's machine.

W2 now covers exactly that: when a turn has no established asset and no history, the guiding question is built from the technician's own words, not from whatever the retriever returned.

And `ct-04`/`gd-05` are now checked in as **must-pass** fixtures — so any future tuning that penalises the coaching style fails by construction. That's the guard against someone re-litigating this from the probe data six months from now.

Unaffected: the self-contradiction (`dc-02`), the 83% footer noise, the invented context, and the oracle fail-open.

Mikecranesync:
## Correction: §9 Q1's recorded answer was overbroad — withdrawn

This PR recorded the answer to *"Is Socratic-by-default the product intent?"* as **"yes, deliberately, for every kind of turn."** That is too broad and has been corrected. It conflicted with three things:

1. **PR #3090** (Cited Technician Turn) says the opposite — direct answer plus one next safe check once context is confirmed and evidence exists.
2. **Merged PR #1685** already established a surface-specific exception: Ignition's single-shot Ask MIRA kiosk answers directly, because the technician cannot take part in a back-and-forth.
3. **This PR contradicted itself** — the main text said Socratic for every turn while its own Definition of Done required direct answers for instructional, documentation, and educational requests.

### Decision of record (§2.2): adaptive dialogue

| Surface | Mode |
|---|---|
| Ignition Ask MIRA, QR, kiosk — single-shot | Direct, complete, cited answer. No forced follow-up. |
| Hub asset / node pages, certified direct connections | Direct cited answer + one next safe check. Never re-confirm a certified asset. |
| Telegram, Slack, conversational web chat | Natural multi-turn conversation. One targeted question when it advances a live diagnosis; direct answer when the answer is supported. |

Intent modulates within a conversational surface: *"what does F004 mean?"* and *"give me the reset procedure"* → direct cited answers; *"why does this conveyor keep stopping?"* with thin evidence → what is known, then one grounded question; *"teach me how to diagnose this circuit"* → deliberate coaching. Safety STOP overrides every mode. **Never withhold a supported answer to satisfy a "never give direct answers" prompt rule.**

### What changed

- **§2.1** — the quiz finding is **reinstated and scoped**, not retired. `ct-04` (a how-to answered with a 3-option quiz while holding a Rockwell citation) is a defect again.
- **§2.2** — rewritten as the adaptive-dialogue decision, with the withdrawal stated explicitly.
- **W2 — restored, not cancelled.** Now two halves: **W2a** picks the mode from surface + intent + evidence and removes *"You never give direct answers"* from `active.yaml`; **W2b** keeps the guiding question grounded in the turn (unchanged). W2a is an engine change → staging gate, separate PR from the probe battery.
- **Rubric** — new dimension **6b Mode appropriateness**, scored separately from 6 Question relevance so neither can be tuned into the other. 6 grades a question that was asked; 6b grades whether asking was right.
- **Fixture policy corrected.** The old text made `ct-04`/`gd-05` **must-pass**, which would have hard-coded the defect into the grader. `ct-04` → must-fail on 6b. `gd-05` → re-label from its actual transcript. **New must-pass fixture required:** a conversational live-diagnosis turn with incomplete evidence that asks exactly one grounded question — without it, 6b is satisfiable by never asking anything, which is the same failure reversed.
- **G1/G3, non-goals, principle 3, P1, risk rows** — reworded to the adaptive policy.

### Preserved unchanged

Question-relevance, invented-context, internal-consistency, citation-footer, and all grader mutation-testing work. **22 detector tests still pass.** The `ct-04` MUST_NOT_FIRE guard stays — those detectors are not the layer that catches mode; `invented_topic` correctly says the vendor came from the technician's own words. Its comment no longer calls the reply "correct behavior."

### Listed, deliberately not changed

Three places still encode the absolute rule. Each moves a scoring baseline, so each needs its own before/after rather than riding along in a PRD correction:

| Location | Conflict |
|---|---|
| `tests/scoring/llm_judge.py:127` | `GSD COMPLIANCE (weight 0.25)` rewards asking with no check on whether asking was appropriate — a correct direct answer to a how-to is penalised |
| `mira-bots/prompts/diagnose/active.yaml` | the *"You never give direct answers"* rule itself (W2a) |
| `tests/eval/synthetic_user.py:4` | synthetic user always expects a question; needs a direct-answer persona to simulate single-shot turns |

Ordering: `active.yaml` first, re-baseline, then reconcile the judge weight.

**Not merging this PR** — it stays open per the correction.

